PR #5, opened 2021-10-09
Fix [:listener, :start] telemetry event

dougw-bc:
The `listener_start` telemetry event was failing because of the `{:ok, port}` tuple response from `transport_module.listen_port(listener_socket) `

- fixed by just grabbing second element from tuple (the actual port number)

### DIFF
--- a/lib/thousand_island/listener.ex
+++ b/lib/thousand_island/listener.ex
@@ -28,16 +28,12 @@ defmodule ThousandIsland.Listener do
       }) do
     case transport_module.listen(port, transport_opts) do
       {:ok, listener_socket} ->
-        :telemetry.execute(
-          [:listener, :start],
-          %{
-            port: transport_module.listen_port(listener_socket) |> elem(1)
-          },
-          %{
-            transport_module: transport_module,
-            transport_opts: transport_opts
-          }
-        )
+        {:ok, port} = transport_module.listen_port(listener_socket)
+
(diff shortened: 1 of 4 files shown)


mtrudel:
Thanks for the PR! Are you in a hurry to see this pushed to hex? I’ve had ‘cover telemetry in tests’ on my list for a while and here’s as good of a place to do it as anywhere (obviously you’re welcome to do it if you’d like, or I’ll get to it in the next few days).

mtrudel:
@dougw-bc can you elaborate on what was failing here? I see that the `[:listener, :start]` telemetry measurement returns a tuple as the port (which is odd & wrong and will be fixed, but shouldn't cause any crashes, unless you were trying to `puts` any of it, maybe?). 

From a testing perspective, it turns out I'd actually added tests covering this exact telemetry call already, but don't actually match on any details in the measurement. I'll tighten those tests up for this and other telemetry calls and land it here, likely later today. In the meantime it would be good to know if you were seeing any actual crashes beyond this.

dougw-bc:
Here is the error msg the pr fixes. 

```
14:32:57.469 [error] Handler "Elixir.ThousandIsland.Logger.info" has failed and has been detached. Class=:error
Reason=%Protocol.UndefinedError{
  description: "",
  protocol: String.Chars,
  value: {:ok, 9000}
}
...
```

PR #9, opened 2022-01-08
Add callback to pass GenServer options to handlers

lucacorti:
Currently if you want to pass some option to the GenServer you need to implement your own GenServer, which is fairly more complicated than just the `use ThousandIsland.Handler` way.

Personally, I need to register handlers using `GenServer.start_link/3` and passing `name: {:via, Registry, {MyRegistry, term}}` to be able to later retrieve handler processes at runtime to be able to call them.

This looked pretty straightforward so I simply drafted the PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,7 @@ thousand_island-*.tar
 
 # Ignore dialyzer caches
 /priv/plts/
+
+
+# ElixirLS
+/.elixir_ls
--- a/lib/thousand_island.ex
+++ b/lib/thousand_island.ex
@@ -37,7 +37,7 @@ defmodule ThousandIsland do
 
   ## Starting a Thousand Island Server
 
-  A typical use of `ThousandIsland` might look like the following: 
+  A typical use of `ThousandIsland` might look like the following:
 
   ```elixir
   defmodule MyApp.Supervisor do
@@ -81,7 +81,7 @@ defmodule ThousandIsland do
 
   ## Logging & Telemetry
 
-  As a low-level library, Thousand Island purposely does not do any inline 
+  As a low-level library, Thousand Island purposely does not do any inline
   logging of any kind. The `ThousandIsland.Logging` module defines a number of
   functions to aid in tracing connections at various log levels, and such logging
   can be dynamically enabled and disabled against an already running server. This
@@ -106,8 +106,8 @@ defmodule ThousandIsland do
   * `[:socket, :shutdown]`: Emitted whenever a `ThousandIsland.Socket.shutdown/2` call completes.
   * `[:socket, :close]`: Emitted whenever a `ThousandIsland.Socket.close/1` call completes.
 
-  Where meaurements indicate a time duration they are are expressed in `System` 
-  `:native` units for performance reasons. They can be conveted to any desired 
+  Where meaurements indicate a time duration they are are expressed in `System`
+  `:native` units for performance reasons. They can be conveted to any desired
   time unit via `System.convert_time_unit/3`.
   """
 
@@ -116,7 +116,7 @@ defmodule ThousandIsland do
 
   * `handler_module`: The name of the module used to handle connections to this server.
   The module is expected to implement the `ThousandIsland.Handler` behaviour. Required.
-  * `handler_options`: A term which is passed as the initial state value to 
+  * `handler_options`: A term which is passed as the initial state value to
   `c:ThousandIsland.Handler.handle_connection/2` calls. Optional, defaulting to nil.
   * `port`: The TCP port number to listen on. If not specified this defaults to 4000.
   If a port number of `0` is given, the server will dynamically assign a port number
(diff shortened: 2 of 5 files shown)


lucacorti:
I realized shortly after that this makes little sense unless you also pass in the `handler_options`.

mtrudel:
Thanks for the PR! Process naming is actually a bit trickier than this in a networking context, and is already accommodated for in PI's design (though we could do a better job of documenting this). It's tricky because of the way that processes are created in TI. Handler processes are created on a one-off basis for every client connection, and the value that is passed in as handler_options is common between them (it's intended to pass in runtime configuration that is common between all handler processes, such as the pid of a backing process or the name of a higher level module to call into, for example). So, even with an `options` callback as you propose here, there's not enough information to be able to meaningfully name the processes that is being created. You might be able to create names based on timestamps or some other derived value, but such names aren't terribly useful. 

Ideally we would know something of the remote end to be able to name the process (ie: the remote IP and port), but the way things are currently designed we don't know that at the time we spin up the new process (TI's design goes out of its way to not block the accepting process for performance reasons, and if we wanted to have that information in hand at process start time we would need to look it up within the acceptor process).

We *do* however have that information at the time that `handle_connection` is called. Within that call, you have full access to the socket (and indeed, this is only a few lines of code after the first time that the socket becomes usefully accessible). You can look up remote IP / port / any other value you would like within this function, and use these values to construct a useful call to the `Registry.register/3` function. Note moreover that since the `handle_connection` call is made within the handler process itself, the ergonomics of doing this are very straightforward.

This approach also avoids a race condition with registering process names via `GenServer.start_link`, namely that the process' socket isn't available until after TI has transferred ownership of the socket to the handler process. If you were to register a name against the process in advance of this, it would be possible to send `handle_info` calls to the process before you had access to the socket (in this case, you would get a nil value in the `{socket, state}` tuple). Waiting to register until the `handle_connection` callback avoids this possibility. 

I think this approach should capture all of your requirements; please let me know if you believe otherwise. 

Thanks again for the PR!

lucacorti:
> Thanks for the PR! Process naming is actually a bit trickier than this in a networking context, and is already accommodated for in TI's design (though we could do a better job of documenting this). It's tricky because of the way that processes are created in TI. Handler processes are created on a one-off basis for every client connection, and the value that is passed in as handler_options is common between them (it's intended to pass in runtime configuration that is common between all handler processes, such as the pid of a backing process or the name of a higher level module to call into, for example). So, even with an `options` callback as you propose here, there's not enough information to be able to meaningfully name the processes that is being created. You might be able to create names based on timestamps or some other derived value, but such names aren't terribly useful.
> Ideally we would know something of the remote end to be able to name the process (ie: the remote IP and port), but the way things are currently designed we don't know that at the time we spin up the new process (TI's design goes out of its way to not block the accepting process for performance reasons, and if we wanted to have that information in hand at process start time we would need to look it up within the acceptor process).

Yes, that's what I am doing. I am passing to handle_options the name of a module from which I can retrieve the configuration needed to setup the handler process in handle_continue. It also allows me to retrieve terms I can use to later lookup the process through Registry. In my case, that's all the information I need to actually retrieve the process at a later time.

> We _do_ however have that information at the time that `handle_connection` is called. Within that call, you have full access to the socket (and indeed, this is only a few lines of code after the first time that the socket becomes usefully accessible). You can look up remote IP / port / any other value you would like within this function, and use these values to construct a useful call to the `Registry.register/3` function. Note moreover that since the `handle_connection` call is made within the handler process itself, the ergonomics of doing this are very straightforward.

Indeed that would also fit my use case just fine.
 
> This approach also avoids a race condition with registering process names via `GenServer.start_link`, namely that the process' socket isn't available until after TI has transferred ownership of the socket to the handler process. If you were to register a name against the process in advance of this, it would be possible to send `handle_info` calls to the process before you had access to the socket (in this case, you would get a nil value in the `{socket, state}` tuple). Waiting to register until the `handle_connection` callback avoids this possibility.

Yes, even though you can simply match the handle_info(msg, {nil, state}) case.

Your approach works for me, the only thing is that you could use this to maybe set other GenServer options like debug?

mtrudel:
> Your approach works for me, the only thing is that you could use this to maybe set other GenServer options like debug?

Agreed. I don't think any of the possible third-arg options would conflict with the internal working of TI, so they should all be fine to specify. Instead of adding to the handler API, I would propose adding plumbing for them alongside the `handler_options` flow that exists today. Perhaps named `handler_genserver_options`? At the same time, documentation about best practices to name handler processes should be included, along the lines of what I laid out above.

If you agree with this approach and would like to take this on that would be very much welcome. Otherwise I'm happy to work it up.

lucacorti:
Sure, I'll draft a PR doing what you proposed along with documentation updates. Thanks for taking the time to explain the motivation behind the choices you made.

mtrudel:
Sounds great! I look forward to seeing it!

lucacorti:
I've updated the PR to allow passing `GenServer` options via `handler_opts`. This has the benefit of simplifying the implementation. However one test is failing because they pass just an atom in handler_opts.

We can either restrict handler_opts to be a keyword list, or pass the GenServer options separately. What do you think?

mtrudel:
Thanks for the update @lucacorti!

I definitely prefer option 2 - having `genserver_opts` be an option alongside (and *not* inside) `handler_opts`.

lucacorti:
@mtrudel Updated with genserver_options argument for ThousandIsland.start_link

mtrudel:
Looks great! Merged!

mtrudel:
If you can confirm that the library works for you as intended I'll cut a new release

lucacorti:
I've resorted to your advice so I'm not needing this anymore, but I guess it does what we disussed 😅

Only thing that could be improved might be to simplify it by passing `ServerConfig` directly to the handler to simplify the interface, but I guess it is fine as it is.

Thanks!

PR #10, opened 2022-01-29
Check result of `ThousandIsland.Socket.handshake/1`

moogle19:
Added a check for the `ThousandIsland.Socket.handshake/1` return value, in case it fails (e.g. if the path to the tls cert is messed up).

PS: Sorry for the many changes, but my editor removed the whitespaces at the end of the lines.
The real change is just the `case` statement at the bottom.

### DIFF
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -271,8 +271,6 @@ defmodule ThousandIsland.Handler do
           {:error, reason} ->
             {:stop, reason, {socket, state}}
         end
-
-        {:noreply, {socket, state}, {:continue, :handle_connection}}
       end
 
       # Use a continue pattern here so that we have committed the socket
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -264,7 +264,14 @@ defmodule ThousandIsland.Handler do
           acceptor_id: socket.acceptor_id
         })
 
-        ThousandIsland.Socket.handshake(socket)
+        case ThousandIsland.Socket.handshake(socket) do
+          {:ok, socket} ->
+            {:noreply, {socket, state}, {:continue, :handle_connection}}
+
+          {:error, reason} ->
+            {:stop, reason, {socket, state}}
+        end
+
         {:noreply, {socket, state}, {:continue, :handle_connection}}
       end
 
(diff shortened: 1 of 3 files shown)


mtrudel:
Great PR! I can't believe I missed that in my original workup. 

A couple of requests in advance of merging:

1. Would you be able to clean up the extra whitespace, just to keep the PR tight? (feel free to rebase / push a revert commit as desired; I'll squash this branch down when merging)
2. Could you please cover this in a test (you can induce a busted handshake pretty easily with tls options on the client - likely  by mandating a super small set of ciphers or having the server require a client cert).

lib/thousand_island/handler.ex
```diff
@@ -264,7 +264,14 @@ defmodule ThousandIsland.Handler do
...
-        ThousandIsland.Socket.handshake(socket)
+        case ThousandIsland.Socket.handshake(socket) do
+          {:ok, socket} ->
+            {:noreply, {socket, state}, {:continue, :handle_connection}}
+
+          {:error, reason} ->
+            {:stop, reason, {socket, state}}
+        end
+
         {:noreply, {socket, state}, {:continue, :handle_connection}}
```
mtrudel, line 275: Note that there's still a dangling return here.

ewildgoose:
I also think that the typespec for Socket.handshake is incorrect? It specifies something like {:error, String.t()}, however, looking at the erlang docs for handshake() it defines the return value as :ssl.reason(), which is in turn defined as:

  closed | timeout | error_alert()

I'm wondering if we should use :ssl.reason() or any(), since arguably it's overrideable by a custom transport? Opinion?

mtrudel:
The SSL module has some odd bits in its typespecs, and if memory serves I think I took a bit of a shortcut there and just punted on the error tuple spec. If you want to sort them all out and tighten up our public spec @ewildgoose that would be very welcome work. So long as dialyzer passes I'm happy.

moogle19:
@mtrudel 
I added some test, but I am not really happy with using `capture_log` for it. 
But I am not quite sure how to test a failure in a single handler without some hacky stuff like `capture_log` or (ab)use the `TelemetryCollector` to collect the `[:handler, :error]`. 

Do you have any other ideas how to extract the error value from a single failing handler?

moogle19:
Okay, I think I found an acceptable solution by sending the error result to the test process and checking it via `assert_received`.
If you have a better solution, please let me know!

mtrudel:
Solution looks good! Just waiting on CI.

PR #11, opened 2022-01-31
support Socket.getopts/2

ewildgoose:
We already have support for Socket.setopts/2, so also add support
for getopts/2 to read socket options

Note: also correcting the type specifications for {:error, error}
results, which are defined as being Posix error results in general,
however, the SSL app defines it's errors as any(), so we go
with the widest definition

### DIFF
--- a/lib/thousand_island/transport.ex
+++ b/lib/thousand_island/transport.ex
@@ -24,8 +24,11 @@ defmodule ThousandIsland.Transport do
   @typedoc "Connection statistics for a given socket"
   @type socket_stats() :: {:ok, [{:inet.stat_option(), integer()}]} | {:error, :inet.posix()}
 
-  @typedoc "Options which can be set on a socket via setopts/2"
-  @type socket_options() :: [:inet.socket_setopt()]
+  @typedoc "Options which can be set on a socket via setopts/2 (or returned from getopts/1)"
+  @type socket_get_options() :: [:inet.socket_getopt()]
+
+  @typedoc "Options which can be set on a socket via setopts/2 (or returned from getopts/1)"
+  @type socket_set_options() :: [:inet.socket_setopt()]
 
   @typedoc "The direction in which to shutdown a connection in advance of closing it"
   @type way() :: :read | :write | :read_write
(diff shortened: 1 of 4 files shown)


mtrudel:
Thanks for the PR! Looks good; just waiting on CI to see if dialyzer turns up anything.

mtrudel:
CI is failing on formatting. Looks like whatever is causing your format diffs on #12 is happening for you here as well.

mtrudel:
FWIW there's nothing fancy about CI's formatting check. See https://github.com/mtrudel/thousand_island/blob/main/.github/workflows/elixir.yml#L46-L47 for details.

ewildgoose:
I've re-created this patch. I think the issue was an interaction between this and #11 before. One of my lines was longer than 100 chars or something and it was wrapped in #11. However, I'm having trouble not having the two patches relate to each other, so grateful if you could at least commit #11 as this makes everything else easier to read. However, I think this should pass tests now?

lib/thousand_island/transport.ex
```diff
@@ -95,11 +98,17 @@ defmodule ThousandIsland.Transport do
             ) ::
               {:ok, non_neg_integer()} | {:error, String.t()}
 
+  @doc """
+  Gets the given options on the socket.
+  """
+  @callback getopts(socket(), socket_get_options()) ::
+              {:ok, socket_set_options()} | {:error, String.t()}
```
mtrudel, line 105: Should this be socket_get_options?

mtrudel (changes requested):
One inline question + merge conflicts to resolve, then this should be ready to go.

lib/thousand_island/transport.ex
ewildgoose: I don't think so?

Explaining in plain to avoid confusion:

* Input param is a "get_options" type
* Return param is from the "set_options" type

I guess intuitively this makes sense as what you getopts(), can be fed right back into setopts()

Did I get the spec right above??

ewildgoose:
Merge conflict resolved

lib/thousand_island/transport.ex
mtrudel: I think so. Dialyzer seems to agree, so we're good here.

mtrudel:
Thanks for the PR!

PR #12, opened 2022-01-31
Mix format

ewildgoose:
I ran mix format on the lib dir as there were a number of cases of trailing spaces in the docs (just means future patches won't accidentally try to include them).

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ complete description, consult `ThousandIsland.start_link/1`):
 
 * `handler_module`: The name of the module used to handle connections to this server.
 The module is expected to implement the `ThousandIsland.Handler` behaviour. Required.
-* `handler_options`: A term which is passed as the initial state value to 
+* `handler_options`: A term which is passed as the initial state value to
 `c:ThousandIsland.Handler.handle_connection/2` calls. Optional, defaulting to nil.
 * `port`: The TCP port number to listen on. If not specified this defaults to 4000.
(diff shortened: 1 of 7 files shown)


ewildgoose:
If you could please check this and commit. Basically it's just removing dangling whitespace on the ends of lines. I have my editor and git set to auto-remove these, hence it was tricky to contribute to the project without accidentally checking in whitespace changes. (If you use  VS Code then there are some plugins to highlight whitespace and unicode chars, also you can have it auto purge dangling whitespace on a per file type. Git also has similar options, but they are either global or per project, so less control)

Grateful if you can commit this one first as I'm not clear how to submit the other patch (#12) without pulling in pieces of this one at the same time?

mtrudel:
Thanks for the PR! I'm fine merging this so long as mix format / credo continue to be happy. I'm going to take a bit of time and try and figure out why they weren't complaining about this previously, but one way or the other I'll get this in so long as CI is green.

ewildgoose:
I observe that "mix format" doesn't trim trailing whitespace (so I'm incorrect to call the issue "mix format" - I guess the changes happened as I hit save + I did a mix format and was confused which caused the changes).

Anyway, I think its helpful if you can merge these as it's a common setting to have auto whitespace trimming set, either in git or your editor. At present it's just a (minor) barrier to contributing as you need to "git add -i" to avoid sending whitespace changes with a commit. Doesn't affect anything else of course

PR #17, opened 2022-09-25
Loosen telemetry version requirement

mcrumm:
Thanks for all the work that you're doing! When I tried to install your Phoenix fork w/ Bandit in an existing app with lots of dependencies, this was literally my only roadblock. Everything else Just Works™ 🎉 🚫 🤠

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -28,7 +28,7 @@ defmodule ThousandIsland.MixProject do
 
   defp deps() do
     [
-      {:telemetry, "~> 1.0"},
+      {:telemetry, "~> 0.4 or ~> 1.0"},
       {:ex_doc, "~> 0.25", only: [:dev, :test], runtime: false},
       {:dialyxir, "~> 1.0", only: [:dev, :test], runtime: false},
       {:credo, "~> 1.5", only: [:dev, :test], runtime: false}


mtrudel:
Looks good! I'm going to take a quick gander at the possible fallout in the morning but I don't suspect there will be any. To be clear, the issue was around a dependency your existing app had on telemetry and not anything in an otherwise new Phoenix app?

mcrumm:
Correct! We have one older dependency that is pinned to telemetry `0.4.x` and I could probably update it, but most other dependencies are matching on either 0.4 or 1.0 as I've done here. Iirc the API had zero changes.

Thanks for your time!

mtrudel:
Great to hear! Any issues with the functionality in your app moving to bandit? I'm still trying to figure out the best way to solicit issues / social proof from folks

mcrumm:
I've only done a cursory local test so far, but everything appears to be working correctly– my LiveViews are requesting and receiving GraphQL subscription data over WebSockets in the usual way :D

mtrudel:
Looking over https://github.com/beam-telemetry/telemetry/compare/v0.4.3..v1.0.0 you're correct! 

Thanks for the PR!

mtrudel:
Published at 0.5.11, https://github.com/phoenixframework/phoenix/pull/4973's mix.lock also updated

PR #26, opened 2022-11-21
Use proplists in preference to keyword lists

### LINKED ISSUE #25: No way to set `inet6` option for the TCP transport (and by extension, to SSL transport)
I am trying to run Bandit with FD instead of specifying port. The passed FD is IPv6 socket opened by systemd (I am testing `systemd` library with Bandit) and my `:options` in Bandit are set to:

```elixir
[
  port: 0,
  transport_options: [
    :inet6,
    fd: fd,
    exit_on_close: false
  ]
]
```

But it fails, as `ThousandIsland` is using `Keyword` module, which do not accept "raw" options, I cannot use `inet6: true` as it is not accepted by `:gen_tcp` (as it do not use `proplist` module), I cannot use approach from Ranch of `net: :inet6` as it is not supported at all.

mtrudel:
Fixes #25

### DIFF
--- a/lib/thousand_island/server_config.ex
+++ b/lib/thousand_island/server_config.ex
@@ -26,15 +26,19 @@ defmodule ThousandIsland.ServerConfig do
 
   @spec new(ThousandIsland.options()) :: t()
   def new(opts \\ []) do
+    if !:proplists.is_defined(:handler_module, opts),
+      do: raise("No handler_module defined in server configuration")
+
     %__MODULE__{
-      port: Keyword.get(opts, :port, 4000),
-      transport_module: Keyword.get(opts, :transport_module, ThousandIsland.Transports.TCP),
-      transport_opts: Keyword.get(opts, :transport_options, []),
-      handler_module: Keyword.fetch!(opts, :handler_module),
-      handler_opts: Keyword.get(opts, :handler_options, []),
-      genserver_opts: Keyword.get(opts, :genserver_options, []),
-      num_acceptors: Keyword.get(opts, :num_acceptors, 10),
-      read_timeout: Keyword.get(opts, :read_timeout, :infinity)
+      port: :proplists.get_value(:port, opts, 4000),
+      transport_module:
+        :proplists.get_value(:transport_module, opts, ThousandIsland.Transports.TCP),
+      transport_opts: :proplists.get_value(:transport_options, opts, []),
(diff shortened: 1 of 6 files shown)


lib/thousand_island/server_config.ex
```diff
@@ -26,15 +26,19 @@ defmodule ThousandIsland.ServerConfig do
...
-      read_timeout: Keyword.get(opts, :read_timeout, :infinity)
+      port: :proplists.get_value(:port, opts, 4000),
+      transport_module:
+        :proplists.get_value(:transport_module, opts, ThousandIsland.Transports.TCP),
+      transport_opts: :proplists.get_value(:transport_options, opts, []),
+      handler_module: :proplists.get_value(:handler_module, opts),
+      handler_opts: :proplists.get_value(:handler_options, opts, []),
+      genserver_opts: :proplists.get_value(:genserver_options, opts, []),
+      num_acceptors: :proplists.get_value(:num_acceptors, opts, 10),
+      read_timeout: :proplists.get_value(:read_timeout, opts, :infinity)
```
hauleth, line 41: Here `Keyword` was OK, it was not needed to change. Only `transport_options` could be a "proplist", but even there it should not matter much.

lib/thousand_island/transports/ssl.ex
```diff
@@ -58,14 +58,15 @@ defmodule ThousandIsland.Transports.SSL do
       reuseaddr: true
     ]
 
-    resolved_options =
-      default_options |> Keyword.merge(user_options) |> Keyword.merge(@hardcoded_options)
+    resolved_options = @hardcoded_options ++ user_options ++ default_options
 
-    if Keyword.take(resolved_options, [:keyfile, :key]) == [] do
+    if !(:proplists.is_defined(:keyfile, resolved_options) ||
+           :proplists.is_defined(:key, resolved_options)) do
```
hauleth, line 64: I think that `not/1` and `or/2` generally work better when we know that the arguments will be `boolean()`.

lib/thousand_island/server_config.ex
mtrudel: Yeah, I just figured consistency is better than trying to make minimal changes, especially considering that this ends up being a superset of keyword support.

lib/thousand_island/transports/ssl.ex
dmorneau: `unless` may be clearer too

PR #27, opened 2022-11-25
Overhaul telemetry

mtrudel:
Use an opinionated span-based telemetry model.


mtrudel:
The work here is largely blocked by figuring out the best path forward in terms of making work easy for https://github.com/open-telemetry/opentelemetry-erlang-contrib/issues/136 to happen.

There are two main issues I’m struggling with at the moment (these are mostly notes-to-self, but in sharing them I'm hoping folks with more of an opentelemetry understanding may be able to provide guidance / clarity):

1. I’m building up my telemetry events with explicit notions of parent / containing spans (that is, all of the events have  `span_id` metadata indicating which span they occur within, and `:start` events also contain `parent_id` metadata). Most telemetry providers seem to not do this, instead relying on the opentelemetry stack to somehow manage this with process dictionary magic. Am I doing this wrong, and if not, what’s the best way to expose these explicit relationships to opentelemetry?

2. The hierarchy of spans I’ve worked up for Thousand Island & Bandit are as follows:

  * `:listener` span (this exists for the lifetime of the server)
    * `:acceptor` span (this is a span that wraps all the connections sourced from a given acceptor)
      * `:connection` span (this is a per-client connection span)
         * `:request` span (this span will exist in Bandit & represents an HTTP request; there may be many within a connection due to keepalive/HTTP2 streams)
           * all of Phoenix's spans would nest here

  Does this make sense, given that 1. some of these spans may be weeks+ long, and 2. they largely exist beyond a single ‘client request’ or other chunk of work that would typically be analysed

bryannaegele:
* In the hierarchy list, are each of those levels in a different process or just one?
* For the purposes of a single "http" request, it looks like that logical boundary is defined by `request` and would be handled in bandit?

For observability purposes, the higher levels (listener, acceptor, and connection) could have some value in start/stop/exception events for metrics and logging but not for tracing, so you can divorce the events here from a "span" context if you want. Like it's good to have a way to monitor (via event or query) the number of open connections, connection/acceptor churn, etc as general metrics.

For tracing, otel cares about individual requests and not connections. So even in an HTTP2 stream, each request should be traced separately. That said, we want the boundary for the request lifecycle to be as close to first byte in and last byte out at the handler so that we're accounting for as much of the transmission as possible. It sounds like this `request` event would be what an instrumentation library would utilize.

To the question of passing things around processes, the pdict "magic" tools we have for telemetry and processes are what we'd still use rather than baking otel into thousand island/bandit to keep a generic mechanism for other use cases and libraries.

mtrudel:
> * In the hierarchy list, are each of those levels in a different process or just one?
> * For the purposes of a single "http" request, it looks like that logical boundary is defined by `request` and would be handled in bandit?
Yep, they're all in separate processes (as they follow the process model that TI uses).

> For observability purposes, the higher levels (listener, acceptor, and connection) could have some value in start/stop/exception events for metrics and logging but not for tracing, so you can divorce the events here from a "span" context if you want. Like it's good to have a way to monitor (via event or query) the number of open connections, connection/acceptor churn, etc as general metrics.

I'm tracking all the relations anyway, so it's no harm for me to do so (other than the cost of so many timestamps). 

> For tracing, otel cares about individual requests and not connections. So even in an HTTP2 stream, each request should be traced separately. That said, we want the boundary for the request lifecycle to be as close to first byte in and last byte out at the handler so that we're accounting for as much of the transmission as possible. It sounds like this `request` event would be what an instrumentation library would utilize.

Noted! 

> To the question of passing things around processes, the pdict "magic" tools we have for telemetry and processes are what we'd still use rather than baking otel into thousand island/bandit to keep a generic mechanism for other use cases and libraries.

For any of the Bandit-hosted stuff (ie: everything from Bandit down through Phoenix), a single request happens in a single process so this should just work for HTTP related spans. I'll make sure that the Bandit events reflect this (they should end up looking more or less exactly as Cowboy's do, though I'll likely end up tracking some extra events / durations for header reading etc)

PR #36, opened 2023-01-09
Don't error on timeout

moogle19:
Fixes timeout error (like [Bandit#79](https://github.com/mtrudel/bandit/issues/79)).
Since we have `handle_timeout/2` where an error can be logged if necessary, I think we can safely do it like this.

### DIFF
--- a/test/thousand_island/handler_test.exs
+++ b/test/thousand_island/handler_test.exs
@@ -362,6 +362,9 @@ defmodule ThousandIsland.HandlerTest do
         Logger.info("ping_received")
 
         case ThousandIsland.Socket.recv(socket, 0) do
+          {:error, :timeout} ->
+            {:error, {:shutdown, :timeout}, state}
+
           {:error, reason} ->
             {:error, reason, state}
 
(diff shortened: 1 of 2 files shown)


test/thousand_island/handler_test.exs
```diff
@@ -362,6 +362,9 @@ defmodule ThousandIsland.HandlerTest do
         Logger.info("ping_received")
 
         case ThousandIsland.Socket.recv(socket, 0) do
+          {:error, :timeout} ->
+            {:error, {:shutdown, :timeout}, state}
```
mtrudel, line 366: This change suggests that an error return of `{:shutdown, :timeout}` is part of the public API (or at least part of the advertised return values that `handle_data` accepts). I don't think we want to be changing that (and having this change here may confuse people into thinking that it's a supported thing). I wonder if we should be leaving this as-was an using `@tag :capture_log` if needed to keep the tests quiet

moogle19: Yes, I did't thought about that.
I upgraded the PR to handle the `{:error, :timeout}` inside `handle_continuation/2` so the public API isn't affected.

mtrudel:
Superb!

PR #37, opened 2023-01-16
Fix doc typos

jswanner:
Thanks again for the amazing library!

Hopefully the changes to the code examples are what you intended.

### DIFF
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -71,7 +71,7 @@ defmodule ThousandIsland.Handler do
 
     @impl ThousandIsland.Handler
     def handle_data(msg, _socket, state) do
-      send(self(), {:send, msg})
+      IO.puts(msg)
       {:continue, state}
     end
 
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -71,7 +71,7 @@ defmodule ThousandIsland.Handler do
 
     @impl ThousandIsland.Handler
     def handle_data(msg, _socket, state) do
-      IO.puts(msg)
+      send(self(), {:send, msg})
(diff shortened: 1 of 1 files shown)


lib/thousand_island/handler.ex
```diff
@@ -71,7 +71,7 @@ defmodule ThousandIsland.Handler do
 
     @impl ThousandIsland.Handler
     def handle_data(msg, _socket, state) do
-      IO.puts(msg)
+      send(self(), {:send, msg})
```
mtrudel, line 74: Not quite, but I can see the confusion! The `handle_info({:send, msg}, {socket, state})` call below is meant to be called from somewhere else in your application (or even from an `iex` session) - it sends data *to* the connected client. The `handle_data(msg, _socket, state)` function is called by Thousand Island when data is received *from* the connected client. In an actual implementation, this log message would be replaced by something 'real'; the logging here is correct in as far as this is a super simple demo.

lib/thousand_island/handler.ex
```diff
@@ -103,7 +103,7 @@ defmodule ThousandIsland.Handler do
 
     @impl ThousandIsland.Handler
     def handle_connection(socket, state) do
-      %{address: host} = ThousandIsland.Socket.peer_info(socket)
+      %{address: address} = ThousandIsland.Socket.peer_info(socket)
```
mtrudel, line 106: This is correct; good catch!

mtrudel:
I backed out the one incorrect change (and commented on it above to clear up the confusion)Thanks for the PR!

lib/thousand_island/handler.ex
jswanner: Sorry to give you more work to do in reverting this change.

PR #39, opened 2023-01-21
Allow handling messages from other TCP/SSL sockets

dsdshcym:
by matching raw_socket with %ThousandIsland.Socket{}.socket in handle_info/2

so that user can define their own handle_info/2 and handle tcp/ssl messages from other raw_sockets
(for example, a MitM server with a upstream_socket)

@mtrudel I'm not sure which tests/documentations need to be updated, feel free to point direction for me 🙏

### DIFF
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -314,16 +314,28 @@ defmodule ThousandIsland.Handler do
         |> handle_continuation(socket)
       end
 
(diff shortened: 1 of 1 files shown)


mtrudel:
Looks great! I don't think any tests need updating here; to my eye this is a refinement of an existing feature not a new one. Existing tests should cover this. 

Once CI comes back green I'll merge this and cut a point release. 

Thanks for the PR!

mtrudel:
`0.5.16` just released with this fix incorporated!

PR #51, opened 2023-03-27
Quietly handle handshake errors

### LINKED ISSUE #49: GenServer `:tls_alert` error on handshake failure
`ThousandIsland.Handler` shuts down with an unexpected reason when there's a TLS handshake error, and reports this to the error logger:

```
17:01:55.000 [error] GenServer #PID<0.24370.1> terminating
** (stop) {:tls_alert, {:handshake_failure, 'TLS server: In state wait_finished received CLIENT ALERT: Fatal - Handshake Failure\n'}}
Last message: {:thousand_island_ready, %ThousandIsland.Socket{socket: {:sslsocket, {:gen_tcp, #Port<0.581>, :tls_connection, [option_tracker: #PID<0.24058.1>, session_tickets_tracker: :disabled, session_id_tracker: #PID<0.24059.1>]}, [#PID<0.24369.1>, #PID<0.24368.1>]}, transport_module: ThousandIsland.Transports.SSL, read_timeout: 60000, span: %ThousandIsland.Telemetry{span_name: :connection, telemetry_span_context: #Reference<0.4197377238.932970501.69923>, start_time: -576460732548559697, start_metadata: %{parent_telemetry_span_context: #Reference<0.4197377238.932970498.70998>, remote_address: {127, 0, 0, 1}, remote_port: 56792, telemetry_span_context: #Reference<0.4197377238.932970501.69923>}}}}
```

I think TLS handshake failure should be expected and is normal behavior in the client-server negotiation, and not sure it makes sense to shut it down this way. Also, `:ssl` already reports to the OTP logger. In my case I'm testing by accessing the server on a different hostname than what the cert is tied to, and I wouldn't expect anything in Bandit/Thousand Island to report abnormal behavior (I already know what went wrong client side).

I was thinking of adding a conditional in `ThousandIsland.Handler`:

```elixir
      def handle_info({:thousand_island_ready, socket}, {nil, state}) do
        ThousandIsland.Telemetry.span_event(socket.span, :ready)

        case ThousandIsland.Socket.handshake(socket) do
          {:ok, socket} -> {:noreply, {socket, state}, {:continue, :handle_connection}}
          {:error, {:tls_alert, error} = reason} -> {:stop, {:shutdown, reason}, {socket, state}}
          # Or if only the hanshake failure: {:error, {:tls_alert, {:handshake_failure, _} = reason} -> {:stop, {:shutdown, reason}, {socket, state}}
          {:error, reason} -> {:stop, reason, {socket, state}}
        end
      end
```

But maybe this is exposing too much of the underlying transport logic and it should be in `ThousandIsland.Transports.SSL`:

```elixir
def handshake(socket) do
  case :ssl.handshake(socket) do
    {:ok, socket} -> {:ok, socket}
    {:error, {:tls_alert, _error} = reason} -> {:error, {:shutdown, reason}}
    {:error, error} -> {:error, error}
  end
end
```

@mtrudel what do you think would be correct behavior here? I'm erring on the side of the GenServer closing down normally for any failed TLS negotiation, because the client would be aware of it in any case.

mtrudel:
Fixes #49

### DIFF
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -226,7 +226,8 @@ defmodule ThousandIsland.Handler do
   has already been closed by the time this callback is called. The return value is ignored.
 
   In addition to socket level errors, this callback is also called in cases where `handle_connection/2` or `handle_data/3`
-  return a `{:error, reason, state}` tuple.
+  return a `{:error, reason, state}` tuple, or when connection handshaking (typically TLS
+  negotiation) fails.
   """
   @callback handle_error(
               reason :: any(),
@@ -302,7 +303,7 @@ defmodule ThousandIsland.Handler do
 
         case ThousandIsland.Socket.handshake(socket) do
           {:ok, socket} -> {:noreply, {socket, state}, {:continue, :handle_connection}}
-          {:error, reason} -> {:stop, reason, {socket, state}}
+          {:error, reason} -> {:stop, {:shutdown, {:handshake, reason}}, {socket, state}}
         end
       end
 
@@ -367,6 +368,12 @@ defmodule ThousandIsland.Handler do
       def terminate(:shutdown, {socket, state}) do
         __MODULE__.handle_shutdown(socket, state)
         do_socket_close(socket, :shutdown)
+      end
+
+      # Called if the socket encountered an error during handshaking
+      def terminate({:shutdown, {:handshake, reason}}, {socket, state}) do
+        __MODULE__.handle_error(reason, socket, state)
+        do_socket_close(socket, reason)
       end
 
       # Called if the remote end shut down the connection, or if the local end closed the


danschultzer:
This looks great! What about removing the capture_log from `test/thousand_island/server_test.exs:230` to show there's nothing being logged? Probably the same for `test/thousand_island/server_test.exs:201`.

mtrudel:
> This looks great! What about removing the capture_log from `test/thousand_island/server_test.exs:230` to show there's nothing being logged? Probably the same for `test/thousand_island/server_test.exs:201`.

There are things being logged though - the logging from within the :ssl lib.

mtrudel:
Published as `0.6.5`

danschultzer:
> > This looks great! What about removing the capture_log from `test/thousand_island/server_test.exs:230` to show there's nothing being logged? Probably the same for `test/thousand_island/server_test.exs:201`.
> 
> There are things being logged though - the logging from within the :ssl lib.

FWIW those log entries can be eliminated by passing `log_level: :warning` or `log_level: none` to the transport options. The client also needs to same setting when it uses `:ssl`.

Thanks for the releases, everything looks good on my end now! 😄

PR #60, opened 2023-05-24
Enhances some public API management functions

asakura:
* Adds missing type @spec and @impl
* Uses Enum.reduce/3 and Enum.reduce_while/3 for more efficient list traverse
* Implements error handling where absence of value is possible

### DIFF
--- a/lib/thousand_island/acceptor_pool_supervisor.ex
+++ b/lib/thousand_island/acceptor_pool_supervisor.ex
@@ -3,8 +3,7 @@ defmodule ThousandIsland.AcceptorPoolSupervisor do
 
   use Supervisor
 
-  @spec start_link({server_pid :: pid, ThousandIsland.ServerConfig.t()}) ::
-          :ignore | {:ok, pid} | {:error, {:already_started, pid} | {:shutdown, term} | term}
+  @spec start_link({server_pid :: pid, ThousandIsland.ServerConfig.t()}) :: Supervisor.on_start()
   def start_link(arg) do
     Supervisor.start_link(__MODULE__, arg)
   end
--- a/lib/thousand_island/acceptor_pool_supervisor.ex
+++ b/lib/thousand_island/acceptor_pool_supervisor.ex
@@ -3,17 +3,27 @@ defmodule ThousandIsland.AcceptorPoolSupervisor do
 
   use Supervisor
 
+  @spec start_link({server_pid :: pid, ThousandIsland.ServerConfig.t()}) ::
+          :ignore | {:ok, pid} | {:error, {:already_started, pid} | {:shutdown, term} | term}
   def start_link(arg) do
     Supervisor.start_link(__MODULE__, arg)
   end
 
-  def acceptor_supervisor_pids(pid) do
-    pid
+  @spec acceptor_supervisor_pids(Supervisor.supervisor()) :: [pid()]
+  def acceptor_supervisor_pids(supervisor) do
+    supervisor
     |> Supervisor.which_children()
-    |> Enum.map(fn {_, acceptor_pid, _, _} -> acceptor_pid end)
-    |> Enum.filter(&Kernel.is_pid/1)
+    |> Enum.reduce([], fn
+      {_, acceptor_pid, _, _}, acc when is_pid(acceptor_pid) -> [acceptor_pid | acc]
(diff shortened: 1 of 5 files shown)


lib/thousand_island/acceptor_supervisor.ex
```diff
@@ -3,19 +3,30 @@ defmodule ThousandIsland.AcceptorSupervisor do
...
+    |> Supervisor.which_children()
+    |> Enum.reduce_while(nil, fn
+      {:connection_sup, connection_sup_pid, _, _}, _acc when is_pid(connection_sup_pid) ->
+        {:halt, connection_sup_pid}
 
-    connection_sup_pid
+      _, acc ->
+        {:cont, acc}
+    end)
   end
```
moogle19, line 23: Do you have some benchmarks to show it is faster? 
`Enum.find/2` should already be quite fast.

asakura: You're right. In this case these two are equal performance wise. Now I see that it's better to use Enum.find_value/3 here for code clarity. Let me fix that quickly.

lib/thousand_island/server.ex
```diff
@@ -3,28 +3,43 @@ defmodule ThousandIsland.Server do
...
+  def listener_pid(supervisor) do
+    supervisor
+    |> Supervisor.which_children()
+    |> Enum.reduce_while(nil, fn
+      {:listener, listener_pid, _, _}, _acc when is_pid(listener_pid) ->
+        {:halt, listener_pid}
+
+      _, acc ->
+        {:cont, acc}
+    end)
```
moogle19: Should use `find_value/3` here as well.

lib/thousand_island/server.ex
```diff
@@ -3,28 +3,43 @@ defmodule ThousandIsland.Server do
...
+    supervisor
+    |> Supervisor.which_children()
+    |> Enum.reduce_while(nil, fn
+      {:acceptor_pool_supervisor, acceptor_pool_sup_pid, _, _}, _acc
+      when is_pid(acceptor_pool_sup_pid) ->
+        {:halt, acceptor_pool_sup_pid}
+
+      _, acc ->
+        {:cont, acc}
+    end)
```
moogle19: Should use `find_value/3` here as well.

lib/thousand_island/acceptor_supervisor.ex
```diff
@@ -3,19 +3,34 @@ defmodule ThousandIsland.AcceptorSupervisor do
...
-    {_, connection_sup_pid, _, _} =
-      pid
-      |> Supervisor.which_children()
-      |> Enum.find(&Kernel.match?({:connection_sup, _, _, _}, &1))
+  @spec connection_sup_pid(Supervisor.supervisor()) :: pid() | nil
+  def connection_sup_pid(supervisor) do
+    supervisor
+    |> Supervisor.which_children()
+    |> Enum.find_value(
+      nil,
```
moogle19, line 17: You can drop the `nil` since that is the default value for `find_value/3`

lib/thousand_island/acceptor_pool_supervisor.ex
```diff
@@ -3,17 +3,27 @@ defmodule ThousandIsland.AcceptorPoolSupervisor do
 
   use Supervisor
 
+  @spec start_link({server_pid :: pid, ThousandIsland.ServerConfig.t()}) ::
+          :ignore | {:ok, pid} | {:error, {:already_started, pid} | {:shutdown, term} | term}
```
moogle19, line 7: Can we just use `Supervisor.on_start()` here as return typespec?

lib/thousand_island/acceptor_supervisor.ex
```diff
@@ -3,19 +3,34 @@ defmodule ThousandIsland.AcceptorSupervisor do
 
   use Supervisor
 
+  @spec start_link({server_pid :: pid, ThousandIsland.ServerConfig.t()}) ::
+          :ignore | {:ok, pid} | {:error, {:already_started, pid} | {:shutdown, term} | term}
```
moogle19, line 7: `Supervisor.on_start()`? (See above)

asakura:
Patches have been pushed.

PR #61, opened 2023-05-25
Adds missing specs and enables additional checks in Dialyzer

asakura:
Updates for Transport modules coming soon.


mtrudel:
Looks great! Should I wait to merge this until you get the Transport specs in or would you rather do that in a separate PR?

asakura:
I think you should merge it right away because the upcoming PR has some points for discussion. Thanks!

PR #62, opened 2023-05-25
Refines types for Transport behaviour, TCP, and SSL implementations

asakura:
It turns out that some callbacks actually should have a broader union type, than they have at the moment. In many cases that's because of missing error tuple. But there are a few worth mentioning:
* `@type on_handshake()` needs to be extended with `{:ok, socket(), any()}` because `:ssl.handshake/1` might return it (but TCP.handshake/1 always returns `{:ok, socket()}`)
* `@callback peer_info/1` `{:error, :inet.posix()}` because both implementations uses the same function from `:inet` module.

At the moment `Transport` defines callbacks with really generic types mostly to fulfill more specific types defined in transport implementation modules. I don't really know what would be the best industry practice to apply here, so let me know please.

Note about `:ssl` modules: dialyzer cannot look over indirect calls  and that reduces greatly it ability to infer types. ':ssl` is wrapper around `:tls_socket` and `:dtls_socket` which uses `:gen_tcp` and `:gen_udp` as a transport. So I would expect that functions in `:ssl` would have similar types, but they haven't. Feels like limitation of the current language model to me.

Regarding `@type socket_info`: since the user has a control over what options to pass to sockets, it might be that  `:inet.peername/1` and `:inet.sockname/1` return some special values like `{:unspec, <<>>}` or `{:undefined, any()}`. I had to handle them in order to prevent dialyzer of inferring really weird types.

### DIFF
--- a/lib/thousand_island/transport.ex
+++ b/lib/thousand_island/transport.ex
@@ -16,7 +16,12 @@ defmodule ThousandIsland.Transport do
 
   @typedoc "Information about an endpoint (either remote ('peer') or local"
   @type socket_info() :: %{
-          address: :inet.ip_address() | :inet.local_address(),
+          address:
+            :inet.ip_address()
+            | :inet.local_address()
+            | {:local, binary()}
+            | :unspec
+            | {:undefined, any()},
           port: :inet.port_number(),
           ssl_cert: String.t() | nil
         }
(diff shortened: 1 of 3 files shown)


mtrudel:
> At the moment Transport defines callbacks with really generic types mostly to fulfill more specific types defined in transport implementation modules. I don't really know what would be the best industry practice to apply here, so let me know please.

I think we'd want to follow Postel's law & keep the return types as narrow as possible (even if it means that we need to do some per-transport special handling to handle oddball return types from the underlying implementation). I *think* that those types should be based as closely as possible on `:gen_tcp`, mostly because it's basically the prototypical 'socket', but I don't have super strong feelings here. Do you?

asakura:
No, I don't have a strong feelings here too. Keeping the return types as narrow as possible is a good idea indeed. Those can be widened if needed. Let me try it first.

asakura:
The PR has been updated. Let me know what you think.

mtrudel:
Looks great! Merging!

Thanks for the work!

PR #65, opened 2023-06-05
Redesign Telemetry so its implementation details won't leak outside

asakura:


lib/thousand_island/telemetry.ex
```diff
@@ -261,24 +261,120 @@ defmodule ThousandIsland.Telemetry do
...
+  @type measurements :: :telemetry.event_measurements()
+
+  @type metadata :: :telemetry.event_metadata()
+
   @app_name :thousand_island
 
   @doc false
-  @spec start_span(atom(), map(), map()) :: t()
-  def start_span(span_name, measurements \\ %{}, metadata \\ %{}) do
+  @spec start_listener_span(
```
mtrudel, line 309: (This same question applies to all other changes in this vein, which constitutes most of this PR)

The use of generic function signatures here was intentional, to keep the definition of payloads closely tied to where they're emitted (ie: it ought to be the Listener's job to determine which parameters go into measurements and metadata). Unless I'm missing another benefit here, this approach is going to create a bunch of boilerplate parameter passing every time we want to add/remove a parameter from telemetry.

asakura: I've been trying to map what and from where goes to into telemetry while I was studying the library in details. After a while of mapping I've realized that there is no really good way to build such mapping with tools given by Dialyzer. The main limitations are that it doesn't allow of using multiple @spec declaration for the same function (ie leads to huge hard to read type declarations, that's one of the reasons why typespecs are missing in the Handler behavior) and it doesn't support dependencies between between type arguments (ie you can't say that for :listener only a map with those keys are allowed). By this moment I've already noticed the existence of three spans (listener -> acceptor -> connection) which are require only a limited set of facts to be instantiated. And it was possible to cover these with type checker. For instance:

```elixir
        span_meta = %{
          local_address: local_info.address,
          local_port: local_info.port,
          transport_module: server_config.transport_module,
          transport_options: server_config.transport_options
        }
        span = ThousandIsland.Telemetry.start_span(:listener, %{}, span_meta)
```

So is it important in the context of the call how collected facts are mapped onto measurements and metadata maps? Is it knowing that there are measurements and metadata map underneath important? Or it was only important for code which directly uses the :telemetry library.  After a moment it became:

```elixir
          ThousandIsland.Telemetry.start_listener_span(
            local_info.address,
            local_info.port,
            server_config.transport_module,
            server_config.transport_options
          )
```

And now when the function is fed with wrong data the type checker barks. At least in my mind that creates an additional anti-fragility layer which prevents  corrupted data at no run-time cost of traveling beyond the thousand_island library to a consumer which might be not ready to handle such data at all. I hope that I could elaborate on this PR a bit. 

> this approach is going to create a bunch of boilerplate parameter passing every time we want to add/remove a parameter from telemetry.

But realistically, how often is that going to happen? Even if that would happen there is an escape hatch to keep number of arguments under control: 

```elixir
  @type stats :: %{
          required(:send_oct) => non_neg_integer(),
          required(:send_cnt) => non_neg_integer(),
          required(:recv_oct) => non_neg_integer(),
          required(:recv_cnt) => non_neg_integer()
        }

  @doc false
  @spec stop_connection_span(t(), stats(), reason) :: :ok
        when reason: :shutdown | :local_closed | error,
             error: term()
  def stop_connection_span(span, stats, reason) when reason in [:shutdown, :local_closed],
    do: stop_span(span, stats, %{})

  def stop_connection_span(span, stats, reason),
    do: stop_span(span, stats, %{error: reason})
```

mtrudel:
Thanks for the great & well thought out explanation! I understand & appreciate your point, however I'm not convinced that telemetry event payloads should be treated as a typed construct. Every other library that I'm aware of treats measurement and metadata payloads as a `map()` and leaves it at that (though of course, libraries typically document that structure in comments). This was the intent of the current implementation of `ThousandIsland.Telemetry`, where the various functions are typed to receive `map()` (like [this one](https://github.com/mtrudel/thousand_island/blob/main/lib/thousand_island/telemetry.ex#L276)). It would follow from this that there's no need for `ThousandIsland.Telemetry` to define event-specific functions; the generalized start/stop/etc functions there currently should stay. 

I agree that it's perhaps less than ideal to not have the internal telemetry plumbing be typed, but I think it's worth it given both community conventions and the lack of boilerplate that such an approach affords. There's no unquestionably correct answer here; this is really just a judgement call that could reasonably go either way.

As a way forward on this PR, I think the lexical renaming of variables to indicate the type of the span in question (ie: changing `parent_span` to `acceptor_span`) is a good one. The same applies for the type changes to [use telemetry specific types in place of `map()`](https://github.com/mtrudel/thousand_island/pull/65/files#diff-74141eed7d1e8fb8a4bd7cf6bbad879bc56b12e4e049bed4b321c6b500a1e379R301-R304). 

WDYT?

mtrudel:
> Should I keep them?

That's fine by me!

asakura:
@mtrudel The branch has been partially reverted. Let me know if there anything else. Thanks!

mtrudel:
This is great! Thanks!

PR #67, opened 2023-06-13
Pulling the signatures up from OTP from peer_info/1, local_info/1

asakura:
According to conversation in a review comment at https://github.com/mtrudel/bandit/pull/162

### DIFF
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -361,10 +361,10 @@ defmodule ThousandIsland.Handler do
 
       def handle_info(
             {msg, raw_socket, data},
-            {%ThousandIsland.Socket{socket: raw_socket, span: span} = socket, state}
+            {%ThousandIsland.Socket{socket: raw_socket} = socket, state}
           )
           when msg in [:tcp, :ssl] do
-        ThousandIsland.Telemetry.untimed_span_event(span, :async_recv, %{data: data})
+        ThousandIsland.Telemetry.untimed_span_event(socket.span, :async_recv, %{data: data})
 
         __MODULE__.handle_data(data, socket, state)
         |> handle_continuation(socket)
@@ -451,7 +451,7 @@ defmodule ThousandIsland.Handler do
               ThousandIsland.Socket.t(),
               reason :: :shutdown | :local_closed | term()
             ) :: :ok
-      defp do_socket_close(%ThousandIsland.Socket{span: span} = socket, reason) do
+      defp do_socket_close(socket, reason) do
         measurements =
           case ThousandIsland.Socket.getstat(socket) do
             {:ok, stats} ->
@@ -466,7 +466,7 @@ defmodule ThousandIsland.Handler do
         metadata = if reason in [:shutdown, :local_closed], do: %{}, else: %{error: reason}
 
         _ = ThousandIsland.Socket.close(socket)
-        ThousandIsland.Telemetry.stop_span(span, measurements, metadata)
+        ThousandIsland.Telemetry.stop_span(socket.span, measurements, metadata)
       end
 
       # Dialyzer gets confused by handle_continuation being a defp and not a def
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -323,18 +323,21 @@ defmodule ThousandIsland.Handler do
 
       @impl GenServer
       def handle_info(
-            {:thousand_island_ready, raw_socket, server_config, acceptor_span, start_time},
+            {:thousand_island_ready, raw_socket,
+             %ThousandIsland.ServerConfig{transport_module: transport_module} = server_config,
+             acceptor_span, start_time},
             {nil, state}
           ) do
-        socket = ThousandIsland.Socket.new(raw_socket, server_config)
-
         {ip, port} =
-          case ThousandIsland.Socket.peername(socket) do
+          case transport_module.peername(raw_socket) do
             {:ok, remote_info} ->
               remote_info
 
             {:error, reason} ->
-              throw({:stop, {:shutdown, {:peername, reason}}, {socket, state}})
+              # the socket has been prematurely closed by the client, we can't do anything with it
+              # so we just close the socket, stop the GenServer with the error reason and move on.
+              _ = transport_module.close(raw_socket)
+              throw({:stop, {:shutdown, {:premature_conn_closing, reason}}, {raw_socket, state}})
           end
 
         span_meta = %{remote_address: ip, remote_port: port}
@@ -347,7 +350,7 @@ defmodule ThousandIsland.Handler do
             span_meta
           )
 
-        socket = ThousandIsland.Socket.attach_span(socket, connection_span)
+        socket = ThousandIsland.Socket.new(raw_socket, server_config, connection_span)
         ThousandIsland.Telemetry.span_event(connection_span, :ready)
 
         case ThousandIsland.Socket.handshake(socket) do
@@ -422,12 +425,6 @@ defmodule ThousandIsland.Handler do
         do_socket_close(socket, :shutdown)
       end
 
-      # Called if the remote end closed the connection before we could initialize it
-      def terminate({:shutdown, {:peername, reason}}, {socket, state}) do
-        __MODULE__.handle_error(reason, socket, state)
-        do_socket_close(socket, reason)
-      end
-
       # Called if the socket encountered an error during handshaking
       def terminate({:shutdown, {:handshake, reason}}, {socket, state}) do
         __MODULE__.handle_error(reason, socket, state)
@@ -439,6 +436,11 @@ defmodule ThousandIsland.Handler do
       def terminate({:shutdown, reason}, {socket, state}) do
         __MODULE__.handle_close(socket, state)
         do_socket_close(socket, reason)
+      end
+
+      # Called if the remote end closed the connection before we could initialize it
+      def terminate({:shutdown, {:premature_conn_closing, _reason}}, {raw_socket, state}) do
+        state
       end
 
       # Called if the socket encountered an error. Socket is closed
@@ -466,12 +468,7 @@ defmodule ThousandIsland.Handler do
         metadata = if reason in [:shutdown, :local_closed], do: %{}, else: %{error: reason}
 
         _ = ThousandIsland.Socket.close(socket)
-
-        if not is_nil(span) do
-          _ = ThousandIsland.Telemetry.stop_span(span, measurements, metadata)
-        end
-
-        :ok
+        ThousandIsland.Telemetry.stop_span(span, measurements, metadata)
       end
 
       # Dialyzer gets confused by handle_continuation being a defp and not a def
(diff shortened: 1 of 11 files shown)


lib/thousand_island/handler.ex
```diff
@@ -326,8 +326,8 @@ defmodule ThousandIsland.Handler do
             {:thousand_island_ready, raw_socket, server_config, parent_span, start_time},
             {nil, state}
           ) do
-        peer_info = server_config.transport_module.peer_info(raw_socket)
-        span_meta = %{remote_address: peer_info.address, remote_port: peer_info.port}
+        {:ok, {ip, port}} = server_config.transport_module.peer_info(raw_socket)
```
asakura, line 329: Should be an exception be raised if non `:ok` tuple returned?

lib/thousand_island/transports/ssl.ex
```diff
@@ -143,46 +143,14 @@ defmodule ThousandIsland.Transports.SSL do
...
+  defdelegate local_info(socket), to: :ssl, as: :sockname
 
   # :ssl.peername/1's typespec is incorrect
   @dialyzer {:no_match, peer_info: 1}
 
   @impl ThousandIsland.Transport
   @spec peer_info(socket()) :: ThousandIsland.Transport.on_peer_info()
-  def peer_info(socket) do
-    cert =
-      case :ssl.peercert(socket) do
```
asakura, line 168: Do you have a preference on how to name a new callback for retrieving peer's certificate? `peer_cert/1`?

lib/thousand_island/handler.ex
mtrudel: I think so. Something along the lines of 'remote end closed the connection before we could initialize it' or something. 

We should probably DRY this up and add it to all the places where we grab peer_info

lib/thousand_island/transports/ssl.ex
asakura: I don't see a reason not to do :)

ryanwinchester: But then you're mixing naming conventions. It's a thing people make fun of PHP for.

What if you kept the same name as the lower libraries but in the existing style/convention like `peer_name`, `sock_name`, and `peer_cert`?

lib/thousand_island/handler.ex
```diff
@@ -337,22 +347,23 @@ defmodule ThousandIsland.Handler do
             span_meta
           )
 
-        socket = ThousandIsland.Socket.new(raw_socket, server_config, connection_span)
-
+        socket = ThousandIsland.Socket.attach_span(socket, connection_span)
```
asakura, line 350: I don't really like attaching span to a Socket struct after, because that changes type of Socket struct: span is nil or Telemetry.t(). Dialyzer ignores that, but Gradient start barking on function from Socket module because they require to be non-nil.

@mtrudel What do you think?

mtrudel: Yeah, I think this is why I reached into the transport_module directly to build the peer info in the first place. I think that approach makes more sense unless there's a reason it needs to be this way?

mtrudel:
Other than the one issue above is this good to go now? We'll want to get this and https://github.com/mtrudel/bandit/pull/167 over in the same moment, with version bumps as appropriate.

lib/thousand_island/handler.ex
asakura: That was the reason - termination:
```elixir
...
            {:error, reason} ->
              throw({:stop, {:shutdown, {:peername, reason}}, {socket, state}})

...

      # Called if the remote end closed the connection before we could initialize it
      def terminate({:shutdown, {:peername, reason}}, {socket, state}) do
        __MODULE__.handle_error(reason, socket, state)
        do_socket_close(socket, reason)
      end
```

Socket needs to be properly closed (I assume) out the server side as well even if the connection was closed by the client. `do_socket_close/2` expect wrapped socket in to the Socket struct.

```elixir
      defp do_socket_close(%ThousandIsland.Socket{span: span} = socket, reason) do
```

asakura: Maybe `Socket.new/3` should be responsible for building a span.

asakura: What do you think of following code? Is there simpler way of doing this:

```elixir
      @impl GenServer
      def handle_info(
            {:thousand_island_ready, raw_socket, server_config, acceptor_span, start_time},
            {nil, state}
          ) do
        case ThousandIsland.Socket.new(raw_socket, server_config, acceptor_span, start_time) do
          {:ok, socket} ->
            ThousandIsland.Telemetry.span_event(socket.span, :ready)

            case ThousandIsland.Socket.handshake(socket) do
              {:ok, socket} -> {:noreply, {socket, state}, {:continue, :handle_connection}}
              {:error, reason} -> {:stop, {:shutdown, {:handshake, reason}}, {socket, state}}
            end

          {:error, reason, socket} ->
            {:stop, {:shutdown, {:peername, reason}}, {socket, state}}
        end
      end

  @spec new(
          ThousandIsland.Transport.socket(),
          ThousandIsland.ServerConfig.t(),
          ThousandIsland.Telemetry.t(),
          non_neg_integer()
        ) :: {:ok, t()} | {:error, :inet.posix(), t()}
  def new(raw_socket, server_config, parent_span, start_time) do
    {error?, span_meta_or_reason} =
      case server_config.transport_module.peername(raw_socket) do
        {:ok, {address, port}} ->
          {false, %{remote_address: address, remote_port: port}}

        {:error, reason} ->
          {true, reason}
      end

    span =
      ThousandIsland.Telemetry.start_child_span(
        parent_span,
        :connection,
        %{monotonic_time: start_time},
        (not error? && span_meta_or_reason) || %{}
      )

    socket = %__MODULE__{
      socket: raw_socket,
      transport_module: server_config.transport_module,
      read_timeout: server_config.read_timeout,
      span: span
    }

    if not error? do
      {:ok, socket}
    else
      {:error, span_meta_or_reason, socket}
    end
  end
```

asakura:
> Other than the one issue above is this good to go now?

Except that one - none I can recognize.

lib/thousand_island/handler.ex
mtrudel: The Socket struct was always intended to model the underlying as the user may need to interact with it (that is, it's intended be a user-facing struct first and foremost). To the extent that we use calls against it internally, it's a convenience more than anything. 

If the client closes the connection before we've even had a chance to call `handle_connection`, I don't think we need to worry about signalling the close to the user. It seems to me to be an inconsistent API that might send a `handle_close`callback for a connection that we haven't even called `handle_connection` for yet.

All of which is to say, I don't think we need to be tying ourselves in knots just to accommodate premature closing, certainly not if it makes the code harder to reason about.

asakura: @mtrudel Oh, great to hear that actually, because some of my concerns has immediately gone. The Socket is user facing abstraction and convenience - check. Before the `handle_connection` is called user cannot spoil pristine state of the handler so no need to call the `handle_close` callback to cleanup. 

But there is one left. If a connection dropped and the GenServer stops right after `peername/1`  returned with error, shouldn't we emit a Telemetry event telling about what happened?

lib/thousand_island/handler.ex
```diff
@@ -323,11 +323,24 @@ defmodule ThousandIsland.Handler do
 
       @impl GenServer
       def handle_info(
-            {:thousand_island_ready, raw_socket, server_config, acceptor_span, start_time},
+            {:thousand_island_ready, raw_socket,
+             %ThousandIsland.ServerConfig{transport_module: transport_module} = server_config,
+             acceptor_span, start_time},
```
mtrudel, line 328: As a general style in the library, I prefer to keep pattern matches in function signatures *only* when required for correctness, and to use dot notation in the function body for readability & locality of reference

mtrudel (changes requested):
One last pass on preferring dot access to function signature matching in two places, and I think we're good to go!

asakura:
https://github.com/mtrudel/bandit/pull/167 is ready only needs adjusting of the TI version

lib/thousand_island/socket.ex
```diff
@@ -22,13 +20,19 @@ defmodule ThousandIsland.Socket do
...
-  def new(socket, server_config, span) do
+        ) :: t()
+  def new(
+        raw_socket,
+        %ThousandIsland.ServerConfig{
+          transport_module: transport_module,
+          read_timeout: read_timeout
+        },
+        span
+      ) do
```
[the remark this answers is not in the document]
asakura: Reverted

lib/thousand_island/handler.ex
asakura: Reverted

mtrudel:
Yeah this looks great!

lib/thousand_island/handler.ex
```diff
@@ -326,8 +326,19 @@ defmodule ThousandIsland.Handler do
...
+        {ip, port} =
+          case server_config.transport_module.peername(raw_socket) do
+            {:ok, remote_info} ->
+              remote_info
+
+            {:error, reason} ->
+              # the socket has been prematurely closed by the client, we can't do anything with it
+              # so we just close the socket, stop the GenServer with the error reason and move on.
+              _ = server_config.transport_module.close(raw_socket)
+              throw({:stop, {:shutdown, {:premature_conn_closing, reason}}, {raw_socket, state}})
```
mtrudel, line 338: In 6 or so years of full time Elixir this is like the third time I've actually used `throw` LOL

lib/thousand_island/handler.ex
```diff
@@ -338,21 +349,22 @@ defmodule ThousandIsland.Handler do
...
           {:ok, socket} -> {:noreply, {socket, state}, {:continue, :handle_connection}}
           {:error, reason} -> {:stop, {:shutdown, {:handshake, reason}}, {socket, state}}
         end
+      catch
+        {:stop, _, _} = stop -> stop
       end
 
       def handle_info(
             {msg, raw_socket, data},
-            {%ThousandIsland.Socket{socket: raw_socket} = socket, state}
```
mtrudel, line 352: Ah shoot one more.

lib/thousand_island/handler.ex
```diff
@@ -434,7 +451,7 @@ defmodule ThousandIsland.Handler do
               ThousandIsland.Socket.t(),
               reason :: :shutdown | :local_closed | term()
             ) :: :ok
-      defp do_socket_close(socket, reason) do
```
mtrudel, line 437: ibid.

lib/thousand_island/handler.ex
asakura: You were lucky all this time:

```bash
proprietary_code $ rg throw | wc -l
2506
```

mtrudel: yowza.

lib/thousand_island/handler.ex
asakura: Reverted

lib/thousand_island/handler.ex
asakura: Reverted

PR #72, opened 2023-07-11
handle econnaborted error from OS kernel

dch:
Per chat, these errors are raised when the kernel has terminated a connection,
with prejudice, for example during SSL weakness probing, incomplete 3-way TCP,
handshakes or similar.

It's not possible to get info about the remote socket, to log info, as it's
no longer available from the kernel.

This raised more questions than I'd anticipated, as I'm not really a Telemetry
user yet. Do we need to close the span, per the catch-all handler, or is
it sufficient to emit an event?

### DIFF
--- a/lib/thousand_island/acceptor.ex
+++ b/lib/thousand_island/acceptor.ex
@@ -29,7 +29,7 @@ defmodule ThousandIsland.Acceptor do
       {:error, :too_many_connections} ->
         ThousandIsland.Telemetry.span_event(span, :spawn_error)
 
-      {:error, reason} when reason in [:closed, :einval] ->
+      {:error, reason} when reason in [:closed, :einval, :econnaborted] ->
         ThousandIsland.Telemetry.stop_span(span, %{connections: count})
 
       {:error, reason} ->


mtrudel:
Thanks for the PR!

I'm not sure (in the literal sense; I'm just not convinced one way or the other) that we want to be handling `:econnaborted` specially compared to other errors we could be getting from accept (and which are handled by the final `else` clause below the one you added). 

I believe that the golden path for tracking this in telemetry-land would be to look for `stop` events with the relevant `error: :econnaborted` entry in their metadata (but that's not an expert opinion at all). What are the benefits to doing this via a dedicated event?

(Of note, in reviewing this PR I realized that `:spawn_error` likely doesn't need to be its own event either; I believe it should also be handled as described above. So basically, the theme of this review is **should we be giving some errors their own events as a matter of policy?**  in the large, not just for `:econnaborted` specifically. WDYT?

dch:
+1 for folding them all into a single thing "under the carpet". From my side, they are very easy to find now and distinguish, as I can correlate them with SSL handshake failures in haproxy. I guess most other LB tools/services will have similar logging.

That said, I could see a case for splitting out any error that a user could do something productive about, e.g. increasing beam processes, adding more acceptors, etc.

BTW the more I dig into 1000island and bandit the more I enjoy it. Such well documented and straightforwards code to use. Thank-you. :rainbow: :cloud: :1st_place_medal: :mango: :orangutan:

lib/thousand_island/acceptor.ex
```diff
@@ -29,6 +29,9 @@ defmodule ThousandIsland.Acceptor do
       {:error, :too_many_connections} ->
         ThousandIsland.Telemetry.span_event(span, :spawn_error)
 
+      {:error, :econnaborted} ->
+        ThousandIsland.Telemetry.span_event(span, :econnaborted)
+
       {:error, reason} when reason in [:closed, :einval] ->
```
mtrudel, line 35: Looking at this with fresh eyes, I wonder if we want to just stick `:econnaborted` into this list and close the connection off 'normally'. To your point, there's nothing for us to do here. WDYT @dch?

dch:
sorry for the slow response, I was Down Under for a few months.

I'm happy with doing that. Let me update, rebase & force push this.

dch:
Plunder what you need for the ChangeLog:

```md
* Handle econnaborted messages from kernel, where the connection has been dropped
  by the Operating System before Erlang & Thousand Island can accept it.

  Thousand Island was unable to accept the connection, as the Operating System
  has already terminated it connection before Thousand Island had a chance to
  accept it. This is usually caused by a client connecting and then immediately
  closing the connection, or the Operating System terminating the connection
  due to a timeout, failed TCP handshake, TLS errors, or similar failures.
```

mtrudel:
Thanks Dave!

PR #80, opened 2023-08-18
Fix typo in port option docs

aaronrenner:
It seems like this is supposed to be `listener_info/1`.


mtrudel:
It's actually obtainable via both mechanisms, depending on if you're working with a Socket struct or not. 

I'm on mobile for the next few days but I'll get to this early next week. 

Thanks for the issue!

mtrudel:
Thanks for the PR @aaronrenner!

PR #86, opened 2023-09-04
feat: allow switching the transport while running

### LINKED ISSUE #84: Upgrading the transport mid-connection
We have a (partial) implementation of PostgreSQL server protocol that I recently ported to ThousandIsland. PostgreSQL protocol has a quirk: it always starts as plain TCP, then, as a first couple of bytes, client may send a request to update to one of the supported encrypted protocols - one of them being SSL. It's then expected of the server to run a SSL handshake and proceed over the socket under the SSL wrapping.

I understand that it is possible to achieve by hand-rolling a handler instead of `use ThousandIsland.Handler`, however it may be a useful feature to allow either returning something akin to a `:switch` tuple you have in Bandit, or just returning an updated socket - since all function on `ThousandIsland.Socket` just use `transport_module` field, it would be enough to allow smth like
```elixir 
def handle_data("magic message", socket, state) do
  socket = ThousandIsland.Transport.Ssl.handshake(socket, additional_ssl_options)
  socket = %{socket | transport_module: ThousandIsland.Transport.Ssl}
  {:continue, {socket, state}} # or {:switch, socket, state, _timeout} to be less conflicting
end
```

Because currently I don't see a way of upgrading the connection within `handle_connection`, and I think it would be a good feature.

I can probably offer up a PR if there is a consensus on the best return value for this

icehaunter:
Closes #84

### DIFF
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -179,6 +179,9 @@ defmodule ThousandIsland.Handler do
           {:continue, state :: term()}
           | {:continue, state :: term(), timeout()}
           | {:continue, state :: term(), {:persistent, timeout()}}
+          | {:switch_transport, {module(), handshake_opts :: [term()]}, state :: term()}
+          | {:switch_transport, {module(), handshake_opts :: [term()]}, state :: term(),
+             timeout()}
           | {:close, state :: term()}
           | {:error, term(), state :: term()}
 
@@ -201,6 +204,14 @@ defmodule ThousandIsland.Handler do
   is received. In order to set a persistent timeout for all future messages (essentially
   overwriting the value of `read_timeout` that was set at server startup), a value of
   `{:persistent, timeout}` may be returned.
+  * Returning `{:switch_transport, {module, opts}, state} will cause Thousand Island to try switching the transport of the
+  current socket. The `module` should be an Elixir module that implements the `ThousandIsland.Transport` behaviour.
+  Thousand Island will call `c:ThousandIsland.Transport.handshake/2` for the given module to upgrade the transport in-place.
+  After a successful handshake Thousand Island will switch the socket to an asynchronous mode, as if `{:continue, state}`
+  was returned.
+  * Returning `{:switch_transport, {module, opts}, state, timeout} is identical to the previous case with the addition of a timeout.
+  After a successful handshake Thousand Island will switch the socket to an asynchronous mode with a timeout, as if
+  `{:continue, state, timeout}` was returned.
   * Returning `{:error, reason, state}` will cause Thousand Island to close the socket & call the `c:handle_error/3` callback to
   allow final cleanup to be done.
   """
@@ -486,6 +497,12 @@ defmodule ThousandIsland.Handler do
             _ = ThousandIsland.Socket.setopts(socket, active: :once)
             {:noreply, {socket, state}, timeout}
 
+          {:switch_transport, {module, handshake_opts}, state} ->
+            handle_switch_continuation(socket, module, handshake_opts, state, socket.read_timeout)
+
+          {:switch_transport, {module, handshake_opts}, state, timeout} ->
+            handle_switch_continuation(socket, module, handshake_opts, state, timeout)
+
           {:close, state} ->
             {:stop, {:shutdown, :local_closed}, {socket, state}}
 
@@ -494,6 +511,18 @@ defmodule ThousandIsland.Handler do
 
           {:error, reason, state} ->
             {:stop, reason, {socket, state}}
+        end
+      end
+
+      defp handle_switch_continuation(socket, module, handshake_opts, state, timeout) do
+        case ThousandIsland.Socket.switch_transport(socket, module, handshake_opts) do
+          {:ok, socket} ->
+            # Upgrade successful, set the socket into active mode again
+            _ = ThousandIsland.Socket.setopts(socket, active: :once)
+            {:noreply, {socket, state}, timeout}
+
+          {:error, reason} ->
+            {:stop, {:shutdown, {:handshake, reason}}, {socket, state}}
         end
       end
     end
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -295,7 +295,6 @@ defmodule ThousandIsland.Handler do
 
   @spec __using__(any) :: Macro.t()
   defmacro __using__(_opts) do
-    # credo:disable-for-next-line Credo.Check.Refactor.LongQuoteBlocks
     quote location: :keep do
       @behaviour ThousandIsland.Handler
 
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -174,14 +174,16 @@ defmodule ThousandIsland.Handler do
   likely that such spans will not behave as expected.
   """
 
+  @typedoc "The possible ways to indicate a timeout when returning values to Thousand Island"
+  @type timeout_options :: timeout() | {:persistent, timeout()}
+
   @typedoc "The value returned by `c:handle_connection/2` and `c:handle_data/3`"
   @type handler_result ::
           {:continue, state :: term()}
-          | {:continue, state :: term(), timeout()}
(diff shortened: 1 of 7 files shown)


icehaunter:
@mtrudel, this PR doesn't currently pass dialyzer with the following error:

```
lib/thousand_island/transports/ssl.ex:82:missing_range
The type specification is missing types returned by function.

Function:
ThousandIsland.Transports.SSL.handshake/1

Type specification return types:
{:error, :closed | :timeout | {:tls_alert, {atom(), string()}}} | {:ok, _}

Missing from spec:

  {:ok, {:sslsocket, _, _},
   %{
     :alpn => binary(),
     :ec_point_formats => [0 | 1 | 2],
     :elliptic_curves => [tuple()],
     :max_frag_enum => 1 | 2 | 3 | 4,
     :next_protocol => binary(),
     :renegotiation_info => binary(),
     :signature_algs => [atom() | {_, _}],
     :sni => string(),
     :srp => binary()
   }}
 ```
 
 reason being that I removed a 3-tuple from the `c:handshake/1` spec return type. That 3-tuple `{:ok, socket, props}` can only be returned by `:ssl.handshake/2` when option `{:handshake, :hello}` is specified and leaves the handshake unfinished. Current code (handler.ex#365) just (correctly) ignores that return value in case matches.
 
 Question: would you rather I relax the types on `c:handshake/1` return type back to include a 3-tuple, or add this line to dialyzer ignore somehow?

File: test/thousand_island/handler_test.exs [hunk 1]
```diff
@@ -136,6 +136,42 @@ defmodule ThousandIsland.HandlerTest do
...
+      use ThousandIsland.Handler
+
+      require Logger
+
+      @impl ThousandIsland.Handler
+      def handle_connection(socket, state) do
+        ThousandIsland.Socket.send(socket, "HELLO")
+
+        {:switch_transport,
+         {ThousandIsland.Transport.SSL,
```
moogle19, line 150: This should be `ThousandIsland.Transports.SSL`

moogle19:
The test currently fails for me with: 

```elixir
1) test handle_connection it should allow upgrading the transport mid-connection (ThousandIsland.HandlerTest)
     test/thousand_island/handler_test.exs:162
     match (=) failed
     code:  assert {:ok, client} = :ssl.connect(client, [cacertfile: Path.join(__DIR__, "../support/ca.pem")], 100)
     left:  {:ok, client}
     right: {:error, {:tls_alert, {:handshake_failure, ~c"TLS client: In state wait_cert_cr at ssl_handshake.erl:2140 generated CLIENT ALERT: Fatal - Handshake Failure\n {bad_cert,hostname_check_failed}"}}}
     stacktrace:
       test/thousand_island/handler_test.exs:168: (test)
```

test/thousand_island/handler_test.exs
icehaunter: Good catch! I wonder how that got through, I did run tests locally before pushing. Anyway, fixing this typo does make the test pass for me. Could you verify that it passes for you as well?

moogle19: Still fails with `{:bad_cert, :hostname_check_failed}`.
Which OTP / Elixir Version and OS are you running?

icehaunter:
```
Erlang/OTP 25 [erts-13.1.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns]

Elixir 1.14.2 (compiled with Erlang/OTP 25)
```
on `Linux 6.4.12-zen1-1-zen x86_64` under ArchLinux, with Elixir/Erlang installed via asdf

moogle19: Ah okay. With OTP 25 it works for me too, but throws this warning: 
```erlang
12:56:37.584 [warning] Description: 'Server authenticity is not verified since certificate path validation is not enabled'
     Reason: 'The option {verify, verify_peer} and one of the options \'cacertfile\' or \'cacerts\' are required to enable this.'
```

OTP 26 changed the default verify to `verify_peer` which seems to be a problem here.

icehaunter: Addressed by setting `verify: :verify_none` in the test

moogle19: The problem here is, that `:ssl.connect/3` doesn't know the hostname it is connecting to. 

A better way to fix the error would be: 
`[cacertfile: Path.join(__DIR__, "../support/ca.pem"), server_name_indication: ~c"localhost"]`

which explicitly sets the SNI for the connection, so the hostname check works.

This probably has to be done by all clients which attempt to upgrade a TCP to a TLS connection (if they check the certificate of the server), so maybe we should add a comment in the docs.

moogle19:
Is a downgrade of a connection theoretically possible (e.g. SSL/TLS -> TCP)? 

With the current implementation switching the SSL/TLS transport with TCP would just fail. 
Either we need to implement it correctly or just don't allow it.

lib/thousand_island/handler.ex
```diff
@@ -179,6 +179,9 @@ defmodule ThousandIsland.Handler do
           {:continue, state :: term()}
           | {:continue, state :: term(), timeout()}
           | {:continue, state :: term(), {:persistent, timeout()}}
+          | {:switch_transport, {module(), handshake_opts :: [term()]}, state :: term()}
```
mtrudel, line 182: We should probably handle `{:persistent, timeout()}` here as well (it could probably all be DRYed up too)

test/thousand_island/handler_test.exs
mtrudel: Yeah, it's inherently gonna be an awkward interface, since all the TLS config stuff that you usually declare at startup needs to be passed as an argument. This is as it should be though, I think.

mtrudel: Not having grokked all of the changes here - do we know why these tests were passing even with the invalid module name? Are we confident that there's no false positive test here?

mtrudel:
> Is a downgrade of a connection theoretically possible (e.g. SSL/TLS -> TCP)?
>
> With the current implementation switching the SSL/TLS transport with TCP would just fail. Either we need to implement it correctly or just don't allow it.

That's a good point. Maybe the new `handshake/2` should be `upgrade/2` instead (still calling into the `:ssl.handshake/2` internal of course). This way transports such as TCP could return `{:error, :not_supported}` or similar (possibly even after inspecting the socket passed in).

test/thousand_island/handler_test.exs
icehaunter: Yes, tests were not passing with the invalid module name. Just opening my own branch on a clean git the tests did fail, I guess a managed to delete a symbol accidentally right before committing. No flaky were observed :)

icehaunter:
@mtrudel @moogle19, all very good points. I think there are two ways to make this interface more generic:
1. Either rename `handshake/2` to `upgrade/2` and then TCP can just disallow "upgrade"
2. or make it properly generic by allowing either the old or the new transport to implement a `upgrade/3` that gets both the old and the new transport modules and makes the decision how and if at all this transition is possible.

(2) feels like a lot of complexity, and even more complexity to make it end-user-extensible. We could maybe do an `Upgradable` protocol, to allow adding arbitrary pairs of transports including existing TCP/SSL? 

I'm not sure I want to pull this into scope though, and I'd rather rename to `upgrade/2` for now. What do you think?

mtrudel:
I think the simple rename to upgrade is actually the better option anyway!

mtrudel:
I'm starting to undertake some related work to this (switching of the handler in the same manner). @icehaunter do you want me to carry this over the line as part of that work? Happy to. LMK

mtrudel:
Updated to `upgrade` naming convention. Ready for review

mtrudel:
@icehaunter, @ryanwinchester could I please get a review on this? Fixin' to merge this and a few other PRs for a release in the next couple of weeks

icehaunter (commented):
@mtrudel sorry for late response, thanks for cleaning this up! I like the upgrade path, and it looks great to me.

ryanwinchester (approved):
Looks good from here

(although `Process.sleep` in tests hurts me emotionally, I can see it's the convention here)

PR #95, opened 2023-10-31
Make handle_continuation public

### LINKED ISSUE #94: A way to align custom and library `GenServer` responses
Hello there 👋

Thank you for such a great library!

There's a tiny itch that I've encountered while while implementing a `ThousandIsland.Handler`. I want to discuss whether it makes sense to solve it the way I did it and whether it makes sense to solve it on the library level.

So, I created a couple of `GenServer.handle_info/cast/call` callbacks while implementing a `ThousandIsland.Handler`. Everything works superb. What's more, those handlers share the logic with `ThousandIsland.Handler.handle_data` aside from one thing -  return values.

I end up with the function that casts `ThousandIsland.Handler.handler_result()` to what `GenServer` handlers expect. Such function exists in the library - `ThousandIsland.Handler.handle_continuation/2`, but it's private.

I wonder whether it would make sense to make the function public? This way, library clients can tap into `GenServer` logic, but align their responses to what `ThousandIsland.Handler` does under the hood with its callbacks. E.g., share the same shutdown methods, or use `socket.read_timeout` as a default value for `GenServer` responses, and such.

Edit: a tiny PR that demonstrates the proposition https://github.com/mtrudel/thousand_island/pull/95

elfenlaid:
Closes #94

The issue is up for discussion 🙇

### DIFF
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -454,31 +454,19 @@ defmodule ThousandIsland.Handler do
         do_socket_close(socket, reason)
       end
 
-      @spec do_socket_close(
-              ThousandIsland.Socket.t(),
-              reason :: :shutdown | :local_closed | term()
-            ) :: :ok
-      defp do_socket_close(socket, reason) do
-        measurements =
-          case ThousandIsland.Socket.getstat(socket) do
-            {:ok, stats} ->
-              stats
-              |> Keyword.take([:send_oct, :send_cnt, :recv_oct, :recv_cnt])
-              |> Enum.into(%{})
+      @doc """
+      This function bridges `ThousandIsland.handler_result()` response values to `GenServer` ones.
 
-            _ ->
-              %{}
-          end
-
-        metadata = if reason in [:shutdown, :local_closed], do: %{}, else: %{error: reason}
-
-        _ = ThousandIsland.Socket.close(socket)
-        ThousandIsland.Telemetry.stop_span(socket.span, measurements, metadata)
-      end
-
-      # Dialyzer gets confused by handle_continuation being a defp and not a def
-      @dialyzer {:no_match, handle_continuation: 2}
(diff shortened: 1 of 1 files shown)


lib/thousand_island/handler.ex
```diff
@@ -507,6 +495,28 @@ defmodule ThousandIsland.Handler do
             end
         end
       end
+
+      @spec do_socket_close(
```
elfenlaid, line 499: I swapped `do_socket_close` and `handle_continuation`, as the latter is now public

lib/thousand_island/handler.ex
```diff
@@ -454,31 +454,19 @@ defmodule ThousandIsland.Handler do
...
-      defp handle_continuation(continuation, socket) do
+      @doc """
+      This function bridges `ThousandIsland.handler_result()` response values to `GenServer` ones.
+
+      The function is useful if the implementation needs to tap into `GenServer.handle_*` methods,
+      but wants to match the returned values to what `ThousandIsland.Handler` returns
+      from `c:handle_connection/2` and `c:handle_data/3` to `GenServer` handlers.
+      """
+      @spec handle_continuation(handler_result(), ThousandIsland.Socket.t()) ::
+              {:noreply, new_state}
```
elfenlaid, line 465: The return type might look weird, I took it from the `GenServer.handle_cast` callback https://github.com/elixir-lang/elixir/blob/7e4fbe657dbf9c3e19e3d2bd6c17cc6d724b4710/lib/elixir/lib/gen_server.ex#L561

elfenlaid:
After going back and forth in the discussion around the ticket https://github.com/mtrudel/thousand_island/issues/94, we decided not to complicate things further.

Instead, we added clear error messages in places where things could go wrong: [Add a check for custom handle_info return types #96](https://github.com/mtrudel/thousand_island/pull/96/files)

PR #97, opened 2023-11-09
Add `supervisor_options` option to specify top-level supervisor options

mtrudel:
I can't believe this wasn't in the library already, but this adds support for specifying top-level options to the supervisor at top of a Thousand Island server instance. Specifically, this allows you to name a server for easier access (see [the docs](https://hexdocs.pm/elixir/1.12/GenServer.html#module-name-registration) for more). An example:

```elixir
iex(1)> ThousandIsland.start_link(handler_module: Echo, supervisor_options: [name: :my_special_name])
{:ok, #PID<0.254.0>}
iex(2)> ThousandIsland.listener_info(:my_special_name)
{:ok, {{0, 0, 0, 0}, 4000}}
```


lib/thousand_island/server.ex
```diff
@@ -5,7 +5,7 @@ defmodule ThousandIsland.Server do
 
   @spec start_link(ThousandIsland.ServerConfig.t()) :: Supervisor.on_start()
   def start_link(%ThousandIsland.ServerConfig{} = config) do
-    Supervisor.start_link(__MODULE__, config)
+    Supervisor.start_link(__MODULE__, config, config.supervisor_options)
```
ryanwinchester, line 8: Not really important but if we don't need these options passed through `init`/etc, maybe this?
```suggestion
    {supervisor_options, config} = Keyword.pop(config, :supervisor_options, [])
    Supervisor.start_link(__MODULE__, config, supervisor_options)
```

mtrudel: Sure!

mtrudel: Enh, we don't do that with any other options, and since config is a full blown `ServerConfig` struct I'm inclined to pass it through unchanged for consistency

ryanwinchester: makes sense

PR #103, opened 2023-12-27
Refine acceptor behaviour on abnormal conditions

mtrudel:
Refines the changes introduced in #72 and #54 to not terminate the acceptor process when abnormal conditions are encountered.

Acceptor tasks have a `transient` restart strategy. Because the exit values when encountering the above conditions otherwise looked like a 'normal' (ie: non-raising) exit from the acceptor task process, the task acceptor supervisor was not restarting the acceptors, meaning that a server that saw a lot of `econnaborted` and/or too many concurrent connections would eventually have all of its acceptor processes die off and would thus stop accepting new connections.


mtrudel:
@dch relevant to your interests

dch:
@mtrudel this looks very good indeed! :rainbow: :christmas_tree: :star_struck: haven't seen an acceptor drop off since

PR #96, opened 2023-11-03
Add a check for custom handle_info return types

elfenlaid:
Follow up on the second part of the solution mentioned in https://github.com/mtrudel/thousand_island/issues/94#issuecomment-1789449398

PR #105, opened 2023-12-31
Move handle_info fallback handler to end of compilation to allow override

### LINKED ISSUE #104: handling non-TI messages within same GenServer fails after PR#96
PR https://github.com/mtrudel/thousand_island/pull/96 adds guard rails to ensure message type matches thousand_island format.
However, it's possible to have additional messages, from outside thousand_island,
destined for the same genserver.

I use this functionality as a tcp proxy, sharing the same GenServer state. This
avoids message-passing overhead that would be introduced between the handler and
a further GenServer if they were separate processes.

- is this a reasonable intended use case for normal TI handlers, or
  should I go for the full custom implementation?

- the guard rails seem currently to prevent this, or am I holding it wrong?

I assume the macro expansions catch the additional handler messages before
my custom ones do, and I'm not clear if this can still be accommodated.

Example cribbed from [messenger.ex](https://github.com/mtrudel/thousand_island/blob/main/examples/messenger.ex) and [handler docs](https://github.com/mtrudel/thousand_island/blob/main/lib/thousand_island/handler.ex#L95-L110)

```elixir
#!/usr/bin/env elixir
Mix.install([{:thousand_island, github: "mtrudel/thousand_island"}])

defmodule Ecco do
  require Logger
  use ThousandIsland.Handler
  alias ThousandIsland.Socket

  @host "httpbin.org"
  @port 80
  @opts [:binary, active: true, nodelay: true]

  @impl ThousandIsland.Handler
  def handle_connection(%Socket{socket: client}, []) do
    {:ok, ip} = String.to_charlist(@host) |> :inet.getaddr(:inet)

    case :gen_tcp.connect(ip, @port, @opts) do
      {:ok, upstream} ->
        Logger.debug("connected to #{@host}:#{@port} via #{inspect(upstream)}")
        {:continue, %{upstream: upstream, client: client}}

      {:error, reason} ->
        Logger.debug("failed to connect to #{@host}:#{@port} because #{reason}")
        {:close, :upstream_connect_failed}
    end
  end

  @impl ThousandIsland.Handler
  def handle_data(data, _socket, state = %{upstream: upstream}) do
    # Send the data to the upstream service
    :gen_tcp.send(upstream, data)
    # Upstream response arrives asynchronously
    {:continue, state}
  end

  @impl GenServer
  def handle_info({:tcp, _port, _msg}, {socket, state}) do
    # just test the shape of the pattern matching in & out
    # ThousandIsland.Socket.send(socket, msg)
    # {:noreply, {socket, state}} also doesn't work
    {:noreply, {socket, state}, socket.read_timeout}
  end
end

# running the server
require Logger

{:ok, _pid} =
  ThousandIsland.start_link(
    port: 1234,
    handler_module: Ecco
  )

Logger.debug("listening on port 1234, curl me please")

unless IEx.started?() do
  Process.sleep(:infinity)
end
```

> you may need to remove ~/.cache/mix/installs/*/*/deps/thousand_island/
> to ensure you get latest main commit as dependency

Then running `curl --trace-ascii - http://localhost:1234` now fails:

```iex
16:22:59.843 [debug] listening on port 1234, curl me please

16:23:02.387 [debug] connected to httpbin.org:80 via #Port<0.9>

16:23:02.800 [error] GenServer #PID<0.499.0> terminating
** (RuntimeError)   The callback's `state` doesn't match the expected `{socket, state}` form.
  Please ensure that you are returning a `{socket, state}` tuple from any
  `GenServer.handle_*` callbacks you have implemented

    /home/dch/.cache/mix/installs/elixir-1.15.7-erts-13.2.2.4/3e8cc75637dec96c7c8ea27f725bf2fe/deps/thousand_island/lib/thousand_island/handler.ex:400: Ecco.handle_info/2
    (stdlib 4.3.1.3) gen_server.erl:1123: :gen_server.try_dispatch/4
    (stdlib 4.3.1.3) gen_server.erl:1200: :gen_server.handle_msg/6
    (stdlib 4.3.1.3) proc_lib.erl:240: :proc_lib.init_p_do_apply/3

Last message: {:tcp, #Port<0.9>, "HTTP/1.1 200 OK\r\n..."}

State: {%ThousandIsland.Socket{socket: #Port<0.7>,
  transport_module: ThousandIsland.Transports.TCP,
  read_timeout: 60000,
  silent_terminate_on_error: false,
  span: %ThousandIsland.Telemetry{...}},
  %{client: #Port<0.7>, upstream: #Port<0.9>}}
```

mtrudel:
Refines the feature added in #96 

Fixes #104 

Would appreciate a review @dch, @elfenlaid, @ryanwinchester, anyone else!

### DIFF
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -396,17 +396,11 @@ defmodule ThousandIsland.Handler do
         {:stop, reason, {socket, state}}
       end
 
-      def handle_info({msg, _raw_socket, _data}, _state) when msg in [:tcp, :ssl] do
-        raise """
-          The callback's `state` doesn't match the expected `{socket, state}` form.
-          Please ensure that you are returning a `{socket, state}` tuple from any
-          `GenServer.handle_*` callbacks you have implemented
-        """
-      end
-
       def handle_info(:timeout, {%ThousandIsland.Socket{} = socket, state}) do
         {:stop, {:shutdown, :timeout}, {socket, state}}
       end
+
+      @before_compile {ThousandIsland.Handler, :add_handle_info_fallback}
 
       # Use a continue pattern here so that we have committed the socket
       # to state in case the `c:handle_connection/2` callback raises an error.
@@ -551,6 +545,18 @@ defmodule ThousandIsland.Handler do
           {:error, reason} ->
             {:stop, {:shutdown, {:upgrade, reason}}, {socket, state}}
         end
+      end
+    end
+  end
+
+  defmacro add_handle_info_fallback(_module) do
+    quote do
+      def handle_info({msg, _raw_socket, _data}, _state) when msg in [:tcp, :ssl] do
+        raise """
+          The callback's `state` doesn't match the expected `{socket, state}` form.
+          Please ensure that you are returning a `{socket, state}` tuple from any
+          `GenServer.handle_*` callbacks you have implemented
+        """
       end
     end
   end
--- a/test/thousand_island/handler_test.exs
+++ b/test/thousand_island/handler_test.exs
@@ -42,6 +42,59 @@ defmodule ThousandIsland.HandlerTest do
       # Ensure that we saw the message displayed by the handle_close callback
       assert messages =~ "Closing with hello"
     end
+
+    defmodule BogusState do
+      use ThousandIsland.Handler
+
+      @impl ThousandIsland.Handler
+      def handle_connection(_socket, state) do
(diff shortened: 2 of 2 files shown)


ryanwinchester (approved):
looks great to me

elfenlaid:
yeap, looks good 🙌

mtrudel:
Released in 1.3.2

dch:
thanks @mtrudel this `@before_compile` trick works perfectly!

PR #111, opened 2024-02-23
Fix config options merging

moogle19:
Currently the `user_options` could overwrite the `@hardcoded_options` but not the `default_options` (`gen_tcp` seems to take the last option if multiple values for 1 key are defined).

This fixes the problem by allowing the `user_options` to overwrite the `default_options` and applying the `@hardcoded_options` last.

PR #112, opened 2024-02-23
Use bare list concatenation to allow non keyword options

mtrudel:
Fixup regression introduced in #111


mtrudel:
Note the failing test here - it seems like maybe `:ssl` uses a different heuristic to parse options than `:gen_tcp` does? But then why does `server_test.exs:422` pass? 

Feel free to poke around and figure out what's going on. I'll likely be able to spend some time on this tomorrow

danschultzer:
I've added made the list unique in #113. I think there should be a test for tcp erlang options as well.

mtrudel:
Closed in favour of #113

PR #113, opened 2024-02-24
Use unique bare list concatenation to allow non keyword options

danschultzer:
Pulling #112  (since I can't push to that existing branch) fixing #111 making the list unique by first occurance. I think it is safer to just pass in unique settings to the erlang modules rather than having duplicates in settings. I don't think the erlang options handling was set up with an expectation that we pass in duplicate settings 😄

### DIFF
--- a/lib/thousand_island/transports/ssl.ex
+++ b/lib/thousand_island/transports/ssl.ex
@@ -57,10 +57,11 @@ defmodule ThousandIsland.Transports.SSL do
       reuseaddr: true
     ]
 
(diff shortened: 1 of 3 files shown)


danschultzer:
This PR is passing in local tests, but I did I see a flakey test error:

Screenshot 2024-02-24 at 6 13 18 AM

I assume it is not related to the changes here.

mtrudel:
Great fix @danschultzer !

mtrudel:
Published as 1.3.5

PR #153, opened 2025-02-16
Support TLS config to use the recently added `certs_keys`

joshk:
This is one part of Phoenix + Bandit supporting the recently(ish) added `certs_keys` config key for the Erlang SSL stack.

https://www.erlang.org/doc/apps/ssl/ssl.html#t:cert_key_conf/0

### DIFF
--- a/lib/thousand_island/transports/ssl.ex
+++ b/lib/thousand_island/transports/ssl.ex
@@ -68,14 +68,14 @@ defmodule ThousandIsland.Transports.SSL do
       )
 
     if not Enum.any?(
-         [:keyfile, :key, :sni_hosts, :sni_fun],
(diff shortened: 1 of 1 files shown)


mtrudel:
Thanks Josh! 

'One part'? I think this should be all that we'd need? Generally we just push config through to the underlying libraries blindly so new flags should 'just work'. The only reason this check is here for certain config is as a hint for users since the underlying system errors are super confusing.

joshk:
This was the other part: https://github.com/elixir-plug/plug/pull/1260

With these two parts merged and shipped, I can publish an early version of my Acme lib which hooks into the `handshake` process.

mtrudel:
Published as 1.3.10

joshk:
❤️ ❤️ ❤️

PR #155, opened 2025-02-21
fix: support `inet_backend` transport option

paulswartz:
While this was available in the spec, it wasn't possible to use. `inet_backend` needs to be the first option, and with the `@hardcoded_options` first that was never possible.


mtrudel:
🙄 bizarre config options never cease to amaze

mtrudel:
Thanks for the PR!

mtrudel:
Published as 1.3.11

PR #174, opened 2025-08-15
Implement multiple listener sockets (`reuseport` support)

NelsonVides:
This just adds one of my favourite features from `ranch` that I noticed missing here, namely, support for multiple listening sockets. One of the best blogs describing what this is good for IMO is this one https://blog.cloudflare.com/the-sad-state-of-linux-socket-balancing/. Arguably, it's quite a low-level optimisation, but there are plenty of occasions when it comes handy.

The logic is that now the acceptor workers get an integer id passed into their arguments, and they use this id to request the listener for a specific socket to accept from. Then the listener might have started only one socket and it will give that one to all the acceptors (the default and existing behaviour), or it might have started many and it will round-robin them across the acceptors.


lib/thousand_island/listener.ex
```diff
@@ -18,50 +18,127 @@ defmodule ThousandIsland.Listener do
...
+           listener_span: listener_span
+         }}
+
       {:error, reason} ->
         {:stop, reason}
     end
   end
 
+  defp start_listen_sockets(%ThousandIsland.ServerConfig{} = server_config) do
+    case server_config.num_listen_sockets do
```
mtrudel, line 54: Other than enforcing resuseport do we need to special case '1' here? I *think* the behaviour would fall out of the same behaviour below? I'm looking for ways to simplify the changes in this file, as it's quite a bit more complex than what was here previously (not a dealbreaker, but I'd love to golf the complexity down as much as possible)

mtrudel:
Sorry, fell through the cracks. This looks generally great! One call out about complexity in listener that it would be great to address if possible. I'd also love to hear of some firsthand validation of this in production if possible as it's a big change. Otherwise this LGTM.

NelsonVides:
Thanks for the CI upgrade! Logic simplified, it was indeed easy and a good change :)

When it comes to using this in prod, I haven't used in with thousand_islands yet, but that I've used with ranch and raw erlang many times. The last one was on this DNS server, DNS is a typical case of a server with extremely abundant but very very short lived requests and this load-balancing shown very nicely in dashboards:
- TCP with ranch: https://github.com/dnsimple/erldns/blob/71426fb8acf91eb48b2a78b5b3a8fab91cebea75/src/listeners/erldns_listeners.erl#L252 with https://github.com/dnsimple/erldns/blob/71426fb8acf91eb48b2a78b5b3a8fab91cebea75/src/listeners/erldns_listeners.erl#L233
- UDP: https://github.com/dnsimple/erldns/blob/71426fb8acf91eb48b2a78b5b3a8fab91cebea75/src/listeners/erldns_listeners.erl#L262

NelsonVides:
Formatter fixed 😬

NelsonVides:
@mtrudel any updates? :)

mtrudel:
Published in 1.4.0

PR #176, opened 2025-08-25
Chunk ssl sendfile

NelsonVides:
Just like I tried to optimise uploads at https://github.com/elixir-plug/plug/pull/1279, now I want to contribute a bit downloading big files over SSL, as I don't want my server to get OOMed :)

The default chunk-size is 8MB, for no special reason TBH. Ranch uses 8MB if no value is provided (https://github.com/ninenines/ranch/blob/d272ca07524a61000324b2b7f7ad1ce9b2dd1487/src/ranch_transport.erl#L126C4-L126C11), but it also allows to pass opts to the underlying `sendfile`. Here `sendfile` takes only four parameters and it just seemed a bit too much to extend the transport interface with a `sendfile` taking an extra argument with a chunk_size (?)

Also tweaked the tests to use files in binary mode, as they also tend to be more memory efficient.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ thousand_island-*.tar
 # Ignore dialyzer caches
 /priv/plts/
 
+/tmp/
 
 # ElixirLS
 /.elixir_ls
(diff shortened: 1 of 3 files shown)


NelsonVides:
@mtrudel dunno if I should ping you in case you haven't seen it? 🙈

NelsonVides:
@mtrudel dunno if I should ping you in case you haven't seen it? 🙈

PR #202, opened 2026-05-20
Separate out network timeout timer from GenServer message timer

mtrudel:
Proposed solution to https://github.com/mtrudel/bandit/issues/577; uses an internal timer (tracked in `read_timer` within the socket struct) to separate out network timeouts from GenServer message timeouts. This has several benefits:

* Network timeouts are now tracked independently of any local mailbox messages which may be received. A busy local mailbox (and so anything happening locally on a schedule) does not reset the network timer; a 10 second read_timeout will ONLY measure 10 seconds of receive idleness.

* GenServer's native OTP semantics around timeouts can be used mostly as in any other GenServer. The only difference is that inbound async network traffic (ie: traffic that would result in a `handle_data` call) resets the GenServer timeout.

### DIFF
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -22,9 +22,7 @@ defmodule ThousandIsland.Handler do
   application as needed. The implementation included in the `use ThousandIsland.Handler` macro uses a `GenServer` structure,
   so you may implement such behaviour via standard `GenServer` patterns. Note that in these cases that state is provided (and
   must be returned) in a `{socket, state}` format, where the second tuple is the same state value that is passed to the various `handle_*` callbacks
-  defined on this behaviour. It also critical to maintain the socket's `read_timeout` value by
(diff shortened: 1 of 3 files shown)


mtrudel:
Of note, there are a few corner cases here:

* The read timeouts are managed entirely by ThousandIsland.Handler's wrappers around handle_data calls. When handle_data completes, the current timer is cancelled / any triggers of it drained, and a new timer is armed and attached to the socket. This means that any synchronous calls you make *within* a handle_* callback have their own separate timeout values as determined by the value of timeout you pass into those calls. Moreover, the timeout timer isn't affected in any way by these calls; it only restarts after completion of the callback, and then only if the current call is a handle_data call (or otherwise one which passes the socket through `ThousandIsland.Handler.handle_continuation/2`). This *should* be pretty manageable logic for folks mixing sync and async receives, but time will tell.

Issue #52, opened 2023-03-28
Unexpected error in accept: :emfile


woylie:
We're getting a lot of these errors:

```
RuntimeError: Unexpected error in accept: :emfile
  File "lib/thousand_island/acceptor.ex", line 27, in ThousandIsland.Acceptor.accept/5
  File "lib/task/supervised.ex", line 89, in Task.Supervised.invoke_mfa/2
  File "proc_lib.erl", line 240, in :proc_lib.init_p_do_apply/3
```

Paired with:

```
** (exit) exited in: GenServer.call(:undefined, :acceptor_info, 5000)
    ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
  File "lib/gen_server.ex", line 1027, in GenServer.call/3
  File "lib/thousand_island/acceptor.ex", line 10, in ThousandIsland.Acceptor.run/1
  File "lib/task/supervised.ex", line 89, in Task.Supervised.invoke_mfa/2
  File "proc_lib.erl", line 240, in :proc_lib.init_p_do_apply/3
```

This happens in a Phoenix application with thousand island 0.6.5 and bandit 0.7.4.

Any indication what could cause this?

moogle19:
Could you pin down which resources wouldn't get freed? 
I would assume Ports / Sockets, but it would be nice to have further informations.

woylie:
>     * What mix of endpoints / transports are you using? HTTP/HTTPS? 1.1/2? WebSockets?

The endpoint serves HTTP 1.1 and WebSockets (LiveView) behind an Envoy proxy that terminates TLS.

>     * What config are you starting Bandit / Cowboy up with?

Only defaults.
 
>     * Do you see any other Bandit / Thousand Island messages in your logs (specifically timeout messages)?
>     * Do the errors happen over time, all at once, after a certain number of connections? Anything we can do to try and repro?
>     * What sort of connection rate (distinct connections / second, requests per connection, etc) does the system in question serve?
>     * Are you able to share methods & charts that demonstrate 'memory usage continuously increasing over time'? Do you happen to have a metric available for total number of VM processes as well? Can you share that?

I'll have to get back to you for that. We're pretty busy at the moment, and I'll be out of office from next week, so it may take a while. The only thing I can give you right now is this 24h error distribution chart:

[...]

PR #206, opened 2026-08-20
Back off on :emfile / :enfile in the accept loop instead of crashing the server

mpettibone:
Note: About ~40 lines of code & ~150 lines of test. Claude helped with tests.

## The problem

The accept loop in `ThousandIsland.Acceptor` treats any accept error it does not explicitly recognize as fatal:

```elixir
{:error, reason} ->
  ThousandIsland.Telemetry.stop_span(span, %{connections: count}, %{error: reason})
  raise "Unexpected error in accept: #{inspect(reason)}"
```

`:emfile` (the process hit its open file descriptor limit) and `:enfile` (the system wide file table is full) both land here. The acceptor task raises, its supervisor restarts it, `accept` immediately returns the same error, and the crash loop begins. Each supervision layer allows 3 restarts in 5 seconds, so the escalation path is short: 4 crashes kill an `AcceptorSupervisor`, 4 of those kill the `AcceptorPoolSupervisor`, and 4 of those kill the whole `ThousandIsland.Server`.

Two measured consequences on unmodified main (`cbc55d1`):

- Sustained exhaustion: with a transport whose `accept/1` persistently returns `{:error, :emfile}`, the entire server tree is dead about 10 ms after startup.
- A brief burst is bad too: because each `AcceptorSupervisor` is `rest_for_one` over the acceptor and its connection supervisor, a burst of just 4 errors on one acceptor kills that supervisor once, and every in-flight connection that acceptor was serving dies with it. Descriptor pressure on a busy node ends up dropping established connections that were doing nothing wrong.

The bitter part is that descriptor exhaustion is usually transient and self healing: some connections close, descriptors free up, and a server that simply waited would recover on its own. Instead the listener disappears at exactly the moment the node is busiest.

## Why it matters

Running out of file descriptors is one of the most common ways a busy BEAM node gets into trouble: a spike in concurrent connections, a low `ulimit -n`, or a descriptor leak elsewhere in the VM. The correct behavior for an acceptor under fd pressure is to throttle and stay alive, not to remove the listener. This exact crash has been reported in the wild against this library: issue #52 ("Unexpected error in accept: :emfile", March 2023) is this code path, and the acceptor's response to `:emfile` is unchanged since then.

## Prior art

Ranch, which Thousand Island is modeled on, handles this explicitly. From `ranch_acceptor.erl` (current master):

```erlang
%% Reduce the accept rate if we run out of FDs.
%% We can't accept anymore anyway, so we might as well wait
%% a little for the situation to resolve itself.
{error, Reason} when Reason =:= emfile; Reason =:= enfile ->
    ranch:log(warning,
        "Ranch acceptor reducing accept rate: out of file descriptors~n",
        [], Logger),
    receive after 100 -> ok end;
```

The Ranch guide documents this as a guarantee: "The accept rate of the listener will be automatically reduced, and a warning message will be logged."

The `accept(2)` man page defines the two errors as the per-process descriptor limit (`EMFILE`) and the system wide limit (`ENFILE`); neither says anything is wrong with the listener socket itself.

## Design

- On `:emfile` or `:enfile`, emit a telemetry event, sleep 100 ms (the same value Ranch uses, kept in a module attribute), and continue the accept loop. The sleep is deliberate backpressure: nothing useful can happen while descriptors are exhausted, and spinning would burn CPU and log volume.
- A telemetry event (`[:thousand_island, :acceptor, :emfile]`) is emitted rather than an inline log, matching the library's stated design of routing everything through telemetry. It is wired into `ThousandIsland.Logger` at the `:error` level, so operators who attach the logger get the same visibility Ranch gives them. The event carries the specific errno in its `error` measurement, since the remediation differs: `:emfile` points at `ulimit -n`, `:enfile` at the system wide table.
- Only `:emfile` and `:enfile` are handled here. The other transient accept errors named by `accept(2)` want a plain retry with no sleep, which is a separate concern proposed in a companion PR, so this change stays tightly scoped.
- The catch all `raise` stays in place for truly unexpected errors, so the library still fails loudly on conditions that really are unrecoverable.

## Regression tests

Two tests in `test/thousand_island/acceptor_emfile_test.exs`, driven by a test transport whose `accept/1` returns from an injected error queue and then falls through to real accepts.

The first opens an echo connection while descriptors are "available", then injects five consecutive `:emfile` / `:enfile` results. Five is chosen deliberately: it exceeds the 3-restarts-in-5-seconds budget, so on unmodified main the acceptor crash loop kills its supervisor and the established connection dies with it. With the fix, the test asserts the established connection keeps echoing through the exhaustion window, no supervisor restarts (same `AcceptorSupervisor` pid before and after), and a new connection is served once descriptors free up. The second test asserts the `:emfile` telemetry event fires with the errno in its `error` measurement.

### DIFF
--- a/test/thousand_island/acceptor_emfile_test.exs
+++ b/test/thousand_island/acceptor_emfile_test.exs
@@ -1,0 +1,152 @@
+defmodule ThousandIsland.AcceptorEmfileTest do
+  use ExUnit.Case, async: false
+
+  use Machete
+
+  # A transport that delegates to gen_tcp but lets the test inject a queue of
+  # accept/1 results via :persistent_term, so we can simulate the OS reporting
+  # file descriptor exhaustion. Once the queue is drained, real accepts resume.
+  defmodule FlakyTransport do
+    @behaviour ThousandIsland.Transport
+
+    def set_accept_results(key, results), do: :persistent_term.put({__MODULE__, key}, results)
+
+    @impl true
+    def listen(port, _opts),
+      do: :gen_tcp.listen(port, [:binary, active: false, reuseaddr: true])
+
+    @impl true
+    def accept(listener_socket) do
+      key = :persistent_term.get({__MODULE__, :key})
+
+      case :persistent_term.get({__MODULE__, key}, []) do
+        [next | rest] ->
+          :persistent_term.put({__MODULE__, key}, rest)
+          # Inject a simulated error without consuming a real connection
+          {:error, next}
+
+        [] ->
+          :gen_tcp.accept(listener_socket)
+      end
+    end
+
+    @impl true
+    def handshake(socket), do: {:ok, socket}
+    @impl true
+    def upgrade(_socket, _opts), do: {:error, :unsupported_upgrade}
+    @impl true
+    def controlling_process(socket, pid), do: :gen_tcp.controlling_process(socket, pid)
+    @impl true
+    def recv(socket, length, timeout), do: :gen_tcp.recv(socket, length, timeout)
+    @impl true
+    def send(socket, data), do: :gen_tcp.send(socket, data)
+    @impl true
+    def sendfile(_s, _f, _o, _l), do: {:error, :not_supported}
+    @impl true
+    def getopts(socket, options), do: :inet.getopts(socket, options)
+    @impl true
+    def setopts(socket, options), do: :inet.setopts(socket, options)
+    @impl true
+    def shutdown(socket, way), do: :gen_tcp.shutdown(socket, way)
+    @impl true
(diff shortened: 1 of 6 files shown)


test/thousand_island/acceptor_emfile_test.exs
```diff
@@ -0,0 +1,152 @@
+defmodule ThousandIsland.AcceptorEmfileTest do
```
mtrudel: Minor nit: name this file acceptor_test

mpettibone: Makes sense. Should be all done now. Thanks again for your reviews and consideration.

PR #208, opened 2026-08-20
Retry the transient network errors accept(2) documents instead of crashing the acceptor

mpettibone:
Note: about ~60 lines of code and ~150 lines of test. Claude helped with the description and tests.

## The problem

The accept loop retries only `:econnaborted` and `:einval`, and treats every other accept error as fatal:

```elixir
{:error, reason} when reason in [:econnaborted, :einval] ->
  ThousandIsland.Telemetry.span_event(span, reason)
  accept(...)

{:error, :closed} ->
  ThousandIsland.Telemetry.stop_span(span, %{connections: count})

{:error, reason} ->
  ThousandIsland.Telemetry.stop_span(span, %{connections: count}, %{error: reason})
  raise "Unexpected error in accept: #{inspect(reason)}"
```

On Linux, `accept` can return a network error that was already pending on the incoming connection: `:enetdown`, `:enetunreach`, `:ehostdown`, `:ehostunreach`, `:enonet`, `:eproto`, `:enoprotoopt`, `:eopnotsupp`, and on some kernels `:etimedout`. Every one of these hits the catch all `raise`. A transient condition on a single inbound connection crashes the acceptor, and a short burst of them escalates: 4 crashes within 5 seconds kill the `AcceptorSupervisor` (dropping that acceptor's in-flight connections, since the connection supervisor lives under it), and sustained bursts walk up the tree toward the whole server, exactly as with descriptor exhaustion.

## Why it matters

These errors are not exotic. They are what the kernel hands back when a connection arrives carrying a pending ICMP error, when a route flaps, or when a probe touches the port in a way the OS surfaces oddly. The existing `:einval` handling was added in 1.3.12 (#162) for exactly this class of thing: HAProxy TCP health checks on FreeBSD, via Bandit issue 460. Today every other error in the set crashes the acceptor instead of being shrugged off.

## Prior art

The Linux `accept(2)` man page is explicit in its error handling note: "Linux accept() (and accept4()) passes already-pending network errors on the new socket as an error code from accept()", and applications should "treat them like EAGAIN by retrying"; for TCP/IP it names "ENETDOWN, EPROTO, ENOPROTOOPT, EHOSTDOWN, ENONET, EHOSTUNREACH, EOPNOTSUPP, and ENETUNREACH". That list is the basis for this change. `:etimedout` is included from the man page's adjacent note that "Various Linux kernels can return other errors such as ENOSR, ESOCKTNOSUPPORT, EPROTONOSUPPORT, ETIMEDOUT"; of those four, only `ETIMEDOUT` is a plausible transient condition of one incoming TCP connection, so the other three deliberately stay on the fail-loud path.

Ranch goes further and retries every accept error that is not `emfile`, `enfile`, or `closed` (a bare `{error, _} -> ok` clause in `ranch_acceptor.erl`). This PR takes the more conservative path of enumerating the documented set, which keeps Thousand Island's fail-loud behavior for anything genuinely unexpected, such as a corrupt listener socket.

## Design

- The transient set is a module attribute, so the list sits in one place and is easy to review against the man page.
- The existing `:econnaborted` / `:einval` clause is untouched, and both keep emitting their existing events. Nothing about current behavior changes; this PR only widens what is survivable.
- The new errnos emit a single `[:thousand_island, :acceptor, :accept_error]` event carrying the specific error in its `error` measurement, following the pattern of `recv_error` and friends. One event to attach to, with the errno available for anyone who wants to discriminate. It is documented and wired into `ThousandIsland.Logger` at the `:error` level.
- The retry has no sleep, unlike the descriptor exhaustion backoff proposed in a companion PR. There is nothing to wait out here; the next accept may well succeed immediately.
- The catch all `raise` stays for anything not in the set.

## Regression tests

Two tests in `test/thousand_island/acceptor_transient_error_test.exs`, using an error-injecting transport. The first opens an echo connection, then injects all ten errnos (the nine new ones plus `:einval`, which also regression-covers the existing path) back to back. Ten consecutive crashes would blow the acceptor's restart budget several times over, so on unmodified main the established connection is killed by the supervisor collapse and the test fails deterministically. With the fix, the established connection keeps echoing through the error burst, the `AcceptorSupervisor` pid is unchanged, and new connections are served afterward. The second test asserts the `:accept_error` event fires carrying the specific errno (`:ehostunreach`) in its measurements.

## Verification

- Both tests fail on unmodified main across seeds and pass on the branch.

No hot path impact: the new clause runs only when `accept` returns one of the listed errors. Fully backward compatible: the existing `:econnaborted` and `:einval` events are untouched, and the only addition to the public surface is the new documented `:accept_error` event.

### DIFF
--- a/test/thousand_island/acceptor_transient_error_test.exs
+++ b/test/thousand_island/acceptor_transient_error_test.exs
@@ -1,0 +1,164 @@
+defmodule ThousandIsland.AcceptorTransientErrorTest do
+  use ExUnit.Case, async: false
+
+  use Machete
+
+  # A transport that injects a queue of accept/1 results (to simulate the OS
+  # returning transient network errors from accept), then falls through to real
+  # gen_tcp accepts.
+  defmodule FlakyTransport do
+    @behaviour ThousandIsland.Transport
+
+    def set_accept_results(key, results), do: :persistent_term.put({__MODULE__, key}, results)
+
+    @impl true
+    def listen(port, _opts),
+      do: :gen_tcp.listen(port, [:binary, active: false, reuseaddr: true])
+
+    @impl true
+    def accept(listener_socket) do
+      key = :persistent_term.get({__MODULE__, :key})
+
+      case :persistent_term.get({__MODULE__, key}, []) do
+        [next | rest] ->
+          :persistent_term.put({__MODULE__, key}, rest)
+          {:error, next}
+
+        [] ->
+          :gen_tcp.accept(listener_socket)
+      end
+    end
+
+    @impl true
+    def handshake(socket), do: {:ok, socket}
+    @impl true
+    def upgrade(_socket, _opts), do: {:error, :unsupported_upgrade}
+    @impl true
+    def controlling_process(socket, pid), do: :gen_tcp.controlling_process(socket, pid)
+    @impl true
+    def recv(socket, length, timeout), do: :gen_tcp.recv(socket, length, timeout)
+    @impl true
+    def send(socket, data), do: :gen_tcp.send(socket, data)
+    @impl true
+    def sendfile(_s, _f, _o, _l), do: {:error, :not_supported}
+    @impl true
+    def getopts(socket, options), do: :inet.getopts(socket, options)
+    @impl true
+    def setopts(socket, options), do: :inet.setopts(socket, options)
+    @impl true
+    def shutdown(socket, way), do: :gen_tcp.shutdown(socket, way)
+    @impl true
+    def close(socket), do: :gen_tcp.close(socket)
+    @impl true
+    def sockname(socket), do: :inet.sockname(socket)
+    @impl true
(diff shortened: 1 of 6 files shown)


test/thousand_island/acceptor_transient_error_test.exs
```diff
@@ -0,0 +1,164 @@
+defmodule ThousandIsland.AcceptorTransientErrorTest do
```
mtrudel: Minor nit: name this file acceptor_test

mpettibone: All done :)

mtrudel:
Can you please resolve the conflicts?

mpettibone:
Ok. This should be good now. Sorry for the delay and thanks for patience.

mtrudel:
Thanks for the PR!

PR #210, opened 2026-08-20
Stop emitting a duplicate connection stop event on handshake and upgr…

mpettibone:
Note: about ~10 lines of code and ~80 lines of test. Claud helped with the desc below and test cases.

## The problem

`ThousandIsland.Socket.handshake/1` stops the connection span itself when the handshake fails:

```elixir
{:error, reason} = err ->
  ThousandIsland.Telemetry.stop_span(socket.span, %{}, %{error: reason})
  err
```

But the caller stops it too. On a failed handshake, `ThousandIsland.Handler` returns `{:stop, {:shutdown, {:handshake, reason}}, ...}`, and its `terminate/2` runs `do_socket_close/2`, which calls `stop_span` again. One connection produces one `:start` event and two `:stop` events. `Socket.upgrade/3` has the identical shape and the identical double emit on a failed upgrade.

Verified on unmodified main by attaching to the connection events and forcing a TLS handshake failure (a client offering only an unsupported cipher): the span emits `:start` once and `:stop` twice.

## Why it matters

`:start` and `:stop` form a span, and consumers rely on the pair being balanced. Anything that matches starts with stops, computes durations, or tracks a live connection gauge by incrementing on start and decrementing on stop is thrown off by the extra stop: double counted handshake errors, mismatched pairs, a gauge that drifts negative. The error metadata for a failed handshake is also reported twice. For a library whose telemetry is its primary introspection surface, the span contract is worth getting exactly right.

## Basis

This one is internal correctness rather than a spec or Ranch comparison. The invariant being restored is Thousand Island's own: a span starts once and stops once.

## Design

Remove the `stop_span` call from the error branches of `Socket.handshake/1` and `Socket.upgrade/3`, and let the handler's terminate path be the single source of the stop event.

Why this direction rather than removing the stop from the handler:

- For the built in `use ThousandIsland.Handler` path, a failed handshake always flows through `terminate/2` into `do_socket_close/2`, which stops the span exactly once, with the connection statistics it gathers and the error in its metadata. That is the authoritative stop. The premature one in `Socket.handshake/1` carries no statistics and races ahead of the real teardown, so it is the redundant one.
- `Socket.handshake/1` and `Socket.upgrade/3` are public functions that handlers built from scratch can call directly, but the `ThousandIsland.Handler` docs already scope the span contract to the built in implementation: "Some of the `:connection` family of telemetry span events are emitted by the `ThousandIsland.Handler` implementation. If you use your own implementation in its place it is likely that such spans will not behave as expected." Removing the duplicate does not change the documented contract for the from-scratch path.

The error return value is unchanged; only the extra telemetry call goes away. A comment is left at each site explaining why the span is intentionally not stopped there, so the next reader does not helpfully add it back.

## Regression tests

Two tests in `test/thousand_island/socket_test.exs`. The first forces a TLS handshake failure and asserts, via `assert_receive` and `refute_receive`, exactly one `:start` and exactly one `:stop` for the failed connection. The second asserts a normal connection also emits exactly one of each, so the fix does not swing the count the other way.

## Verification

- The failed handshake test fails on unmodified main (it observes the second `:stop`) and passes on the branch; the balance test passes on both, as expected.

No hot path impact; it removes a telemetry call on an error path. A consumer somehow relying on the duplicate stop would see one fewer event, but the duplicate was a bug, and the surviving stop is the one carrying the real connection statistics and error metadata.

## Note

This is the smallest and safest change of the five proposals, and it is squarely in the "measurably wrong today" category rather than a judgment call. The telemetry surface is one of the nicer things about this library, which is exactly why the double stop stood out.


test/thousand_island/handshake_telemetry_test.exs
```diff
@@ -0,0 +1,83 @@
+defmodule ThousandIsland.HandshakeTelemetryTest do
```
mtrudel: Minor nit: fold this into socket_test

mpettibone: Done. Folded both tests into socket_test: the balanced start/stop check runs under both the gen_tcp and ssl setups, the failed handshake one is in the ssl block, and they go through TelemetryHelpers so the file stays async-safe

mtrudel: Nice!

mtrudel:
Thanks for the PR!

Issue #141, opened 2024-11-19
changing the socket during transport.handshake


danielspofford:
I have a custom transport whose `handshake/1` callback wishes to mutate the socket, but this does not seem possible right now according to https://github.com/mtrudel/thousand_island/blob/main/lib/thousand_island/socket.ex#L49 always returning the socket as it was prior to the handshake.

Is this an intentional constraint, or something that could be allowed in a future release?

Appreciate this library and your efforts. Hope I'm not missing something obvious or doing something silly.

mtrudel:
Nice catch! I don't think there's any good reason for it to stay the way it is; likely an oversight on my part when I first wrote this.

Candidate fix is up at #142; are you able to validate that that does what you want?

danielspofford:
Weeeee! I can try it right now...

danielspofford:
Failing in a confusing way, will give more information when I understand what's going on.

danielspofford:
@mtrudel confirmed all issues were on my end! Everything is working as expected 🎉.

PR #211, opened 2026-08-20
add a handshake_timeout server option

mpettibone:
Note: ~50 lines of code / ~175 test lines. Claude helped with desc and tests.

## What this adds

A new top-level server option, `handshake_timeout`, bounding how long a freshly accepted connection may take to complete its transport handshake (TLS negotiation, in practice). It defaults to 5000 ms; `:infinity` disables the bound. When the timeout elapses, the connection tears down through the existing handshake-error path: the socket is closed, the `num_connections` slot is reclaimed, and the handler sees `c:handle_error/3` with a reason of `:timeout`.

## Why it matters

Today the SSL transport performs its handshake with `:ssl.handshake(socket)`, which per the OTP docs is equivalent to a call with an infinite timeout. `read_timeout` only begins counting after the handshake returns, and nothing else bounds the window. The consequence: a client that opens a TCP connection to a TLS listener and then goes quiet, or dribbles handshake bytes slowly, holds a connection process, a file descriptor, and one slot of `num_connections` indefinitely. A modest number of such sockets, malicious or just broken, quietly eats capacity; with the default capacity math (100 acceptors x 16,384 connections) the file descriptor limit is usually what actually gives out first.

The OTP ssl documentation calls the risk out directly: "Not setting the timeout makes the server more vulnerable to Denial of Service (DoS) attacks." Ranch has bounded this window since 1.6 with a `handshake_timeout` option defaulting to 5000 ms (confirmed at source: `maps:get(handshake_timeout, TransOpts, 5000)` in ranch_conns_sup), so this is straight competitive parity as well as hardening. It appears there's some history on this subject of servers meet exactly this kind of traffic: the `:econnaborted` handling added to the acceptor in PR #72 was motivated by "SSL weakness probing or incomplete TCP handshakes" observed in the wild.

## Standards and prior art

RFC 8446 does not mandate a server-side handshake timer, but bounding half-open negotiations is standard TLS server hardening, and every mainstream implementation does it: Ranch (`handshake_timeout`, 5000 ms), nginx (`ssl_handshake_timeout`), HAProxy (`timeout client` applies during the handshake). The mechanism here uses only the documented `:ssl.handshake/2` timeout argument, available since OTP 21.

## Design rationale

The plumbing mirrors how `read_timeout` already flows through the codebase: an option on `ThousandIsland`, a field on `ServerConfig`, copied into `HandlerConfig` (the struct acceptors precompute once), carried on the `Socket`, and consumed at the single handshake call site in `Socket.handshake/1`.

The decision worth defending in review is how the timeout reaches transports. `ThousandIsland.Transport.handshake/1` takes only a socket, and custom transports are a documented extension point of this library (the "When Handler Isn't Enough" escape hatch; issues #141 and #144 both involve people building them), so changing that callback's arity would break them at runtime. Instead, `handshake/2` (socket, timeout) is added as an optional callback alongside the existing `handshake/1`: `Socket.handshake/1` prefers `handshake/2` when the transport exports it and falls back to `handshake/1` otherwise, via a one-line `function_exported?/3` check. Both bundled transports implement it (TCP ignores the timeout, since it has no handshake; SSL passes it to `:ssl.handshake/2`), and a dedicated compatibility test exercises a transport that implements only `handshake/1` to lock the fallback in. This is the same optional-callback pattern OTP itself uses to evolve behaviours.

On the default: 5000 ms matches Ranch and the OTP guidance, and makes servers safe out of the box. It is a deliberate behavior change for TLS servers that relied on unbounded handshakes, which is why the docs state it plainly and the suggested PR text below proposes a changelog note in the same spirit as the 1.5.0 timeout-separation note. If the maintainer prefers introducing the option with zero behavior change, flipping the default to `:infinity` is a one-line change and the rest of the PR stands as is.

## Performance

Nothing on the per-byte path. Per connection, the additions are one struct field copy and one `function_exported?/3` check (a BIF, nanoseconds) before the handshake call. The timeout itself is managed inside OTP's ssl app, which already runs a timer-capable state machine per handshake. TCP connections are completely unaffected.

## Tests

Four tests in a new `test/thousand_island/handshake_timeout_test.exs`:

1. Configuration: `handshake_timeout` defaults to 5_000 and accepts `:infinity`.
2. Stalled handshake: a client connects to a TLS listener at the TCP level and never begins a handshake; with `handshake_timeout: 250` the handler's `handle_error/3` receives `:timeout`, `connection_pids/1` returns empty (the slot is reclaimed), and the client observes the close.
3. Happy path: a well-behaved TLS client handshakes and echoes normally under the default timeout, with no option passed.
4. Compatibility: a transport defined in the test file that implements only `handshake/1` (delegating everything else to the TCP transport) serves a full echo round trip with a finite `handshake_timeout` configured, proving third-party transports keep working untouched.


test/thousand_island/handshake_timeout_test.exs
```diff
@@ -0,0 +1,175 @@
+defmodule ThousandIsland.HandshakeTimeoutTest do
```
mtrudel: Minor nit: fold this into socket_test

lib/thousand_island/socket.ex
```diff
@@ -5,13 +5,14 @@ defmodule ThousandIsland.Socket do
   """
 
   @enforce_keys [:socket, :transport_module, :read_timeout, :silent_terminate_on_error, :span]
-  defstruct @enforce_keys ++ [:read_timer]
+  defstruct @enforce_keys ++ [:read_timer, handshake_timeout: 5_000]
```
mtrudel: Do we need to specify 5000 here? won't it flow from the server_config?

lib/thousand_island/handler_config.ex
```diff
@@ -6,12 +6,13 @@ defmodule ThousandIsland.HandlerConfig do
   """
 
   @enforce_keys [:handler_module, :transport_module, :read_timeout, :silent_terminate_on_error]
-  defstruct @enforce_keys
+  defstruct @enforce_keys ++ [handshake_timeout: 5_000]
```
mtrudel: Do we need to specify 5000 here? won't it flow from the server_config?

test/thousand_island/handshake_timeout_test.exs
mpettibone: Done. The behaviour tests are in socket_test now and the default value check moved to config_test next to the existing ServerConfig default tests. The standalone file is gone

lib/thousand_island/socket.ex
mpettibone: You're right, it always flows from server_config (through HandlerConfig.from_server_config and Socket.new). Dropped the default and added it to @enforce_keys instead, so it gets the same treatment as read_timeout.

lib/thousand_island/handler_config.ex
mpettibone: Yep. Same as the socket one, removed. It's an enforced key now like read_timeout, and the 5000 default lives only in ServerConfig.

mpettibone: I'm fixing based on your comments now. I need to rebuild, might need an hour or so.

mtrudel:
Thanks for the PR!

PR #213, opened 2026-08-21
Handler performance: reset the read timer lazily instead of on every message

mpettibone:
#202 gave connections precise network timeout semantics, but the enforcement runs on the hottest path in the library: every `{:continue, ...}` return, meaning every piece of data on every connection, cancels a timer, flushes the mailbox, and arms a fresh one, all to police a deadline that almost never expires. That is per-message cost for a per-idle-period concern, and eprof puts it at roughly 10% of handler CPU on receive-heavy connections, with the mailbox flush scaling with mailbox depth, so the busiest handlers pay the most.

This PR moves the work to where the rare event is: continuations just record the new deadline, and a firing timer checks it, stopping the connection if it has truly expired and re-arming for the remainder if data moved it forward. A timer is armed once per timeout interval instead of once per message. A deadline that moves earlier (a shorter one-shot) still cancels and re-arms eagerly, so nothing ever fires late.

Behaviour is unchanged: `handle_timeout` still fires at last-data-plus-timeout with the same millisecond precision, the one-shot, `{:persistent, timeout}`, and `:infinity` variants all act identically, and the #202 guarantee that local messages never reset the network timeout is untouched. All existing timeout tests pass unmodified. The two tests added here (a one-shot longer than the pending timeout is honored in full, a one-shot `:infinity` disables it) also pass on current main, since they pin behaviour rather than implementation.

On a loopback echo bench (OTP 27.3 / Elixir 1.18.4, medians of interleaved runs) this is worth about +14% on streaming-style traffic and +3% on small keepalive request/response.

`ThousandIsland.Socket` gains two internal fields (`read_deadline`, `timer_deadline`) beside `read_timer`, the same home #202 chose for the timer itself.

### DIFF
--- a/lib/thousand_island/handler.ex
+++ b/lib/thousand_island/handler.ex
@@ -425,7 +425,10 @@ defmodule ThousandIsland.Handler do
       end
 
       def handle_info(:read_timeout, {%ThousandIsland.Socket{} = socket, state}) do
-        {:stop, {:shutdown, :read_timeout}, {socket, state}}
+        case ThousandIsland.Handler.check_read_deadline(socket) do
+          {:timeout, socket} -> {:stop, {:shutdown, :read_timeout}, {socket, state}}
+          {:continue, socket} -> {:noreply, {socket, state}}
+        end
       end
 
       @before_compile {ThousandIsland.Handler, :add_handle_info_fallback}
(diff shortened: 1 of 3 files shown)


mtrudel:
yeah this is smart.

PR #214, opened 2026-08-21
Fix port 0 with num_listen_sockets: bind every listener to the advertised port

mpettibone:
This fixes an edge case in the multiple listener socket support introduced in #174.

When `port: 0` is combined with `num_listen_sockets > 1`, each `listen/2` call receives its own ephemeral port from the operating system, but only the first socket's port is advertised through `listener_info/1`. Connections to the advertised port therefore reach just the first listener socket, and the acceptors assigned to the other sockets wait on ports nothing ever connects to. Reproduced on main with three listen sockets:

```
advertised 33793 | actual ports [{1, 33793}, {2, 41517}, {3, 37121}] | reachable 1/3
```

It subtle but the acceptor round-robin always leaves some acceptors on the first socket, so a `port: 0` multi-socket server still accepts connections and passes the existing tests. It just quietly loses the load-spreading that `num_listen_sockets` exists to provide, along with the acceptors assigned to the unreachable sockets.

The fix opens the first listener with the configured port, reads the concrete port the OS assigned, and binds every remaining listener to that port, so the whole reuseport group is reachable through the advertised address:

```
advertised 46049 | actual ports [{1, 46049}, {2, 46049}, {3, 46049}] | reachable 3/3
```

Fixed nonzero ports flow through the same path with the same behaviour as today. Non-IP listeners (Unix domain sockets, whose sockname has no port to propagate) keep the configured value, so their startup behaviour is also unchanged. As a side effect of the restructure, sockets opened before a later bind failure are now closed rather than left to die with the listener process.

Two tests are added: one asserting that with `port: 0` every listener socket reports the advertised address (fails on main, passes here), and one pinning that Unix domain sockets with `num_listen_sockets > 1` keep failing cleanly with `:eaddrinuse`.

Validation: `mix format --check-formatted`, `mix credo --strict`, and `mix dialyzer` are clean, and the full suite passes.

### DIFF
--- a/lib/thousand_island/listener.ex
+++ b/lib/thousand_island/listener.ex
@@ -53,37 +53,51 @@ defmodule ThousandIsland.Listener do
   end
 
   defp start_listen_sockets(%ThousandIsland.ServerConfig{} = server_config) do
-    num_sockets = server_config.num_listen_sockets
+    with {:ok, first_socket} <-
+           server_config.transport_module.listen(
+             server_config.port,
+             server_config.transport_options
+           ) do
+      sockets = [{1, first_socket}]
 
-    sockets =
-      for socket_id <- 1..num_sockets do
-        case server_config.transport_module.listen(
-               server_config.port,
-               server_config.transport_options
-             ) do
-          {:ok, socket} -> {socket_id, socket}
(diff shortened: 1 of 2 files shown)


mtrudel:
Good catch

PR #224, opened 2026-08-26
Docs: Make the HTTP hello-world example follow RFC 1945

mpettibone:
Note: I asked Claude to specifically look for spec non-conformance. It found below. Claude helped write the desc, test case and identification of the PR.

## Summary

Replace the intentionally noncompliant one-chunk HTTP responder with a small, self-contained origin server implementing a conservative RFC 1945 subset. It now accumulates fragmented TCP input, parses a complete request, supports HTTP/0.9 `GET`, implements HTTP/1.0 `GET` and `HEAD`, returns meaningful errors, emits entity metadata, and closes after one response. The module explicitly describes itself as a subset rather than claiming to be a general HTTP parser.

## Non-conformance

**The rule.** The example declares itself an HTTP/1.0 server, and RFC 1945 defines what that means on the wire: parse the Request-Line and headers before responding ([section 4.1](https://www.rfc-editor.org/rfc/rfc1945.html#section-4.1) distinguishes simple and full messages, [section 5](https://www.rfc-editor.org/rfc/rfc1945.html#section-5) gives the request grammar, [section 4.2](https://www.rfc-editor.org/rfc/rfc1945.html#section-4.2) constrains field values, and [section 5.1.2](https://www.rfc-editor.org/rfc/rfc1945.html#section-5.1.2) requires an `absoluteURI` or `abs_path` target per [RFC 1808 section 2.2](https://www.rfc-editor.org/rfc/rfc1808.html#section-2.2)); answer an HTTP/0.9 Simple-Request with a Simple-Response; send no entity body for `HEAD` ([section 8.2](https://www.rfc-editor.org/rfc/rfc1945.html#section-8.2)); and use `400` and `501` for malformed and unimplemented requests ([sections 9.4 and 9.5](https://www.rfc-editor.org/rfc/rfc1945.html#section-9.4)). Response metadata (`Date`, `Content-Type`, `Content-Length`, [sections 10.4 through 10.6](https://www.rfc-editor.org/rfc/rfc1945.html#section-10.4)) and close-delimited bodies ([section 7.2.2](https://www.rfc-editor.org/rfc/rfc1945.html#section-7.2.2)) are recommendation-level.

**What the example does today.** `examples/http_hello_world.ex` sends one complete `200 OK` HTTP/1.0 response to the first TCP chunk it receives, whatever that chunk contains.

**What goes wrong,** point by point against the RFC:

- it never parses the Request-Line or waits for the header terminator, so a request split across TCP segments is answered before it arrives;
- an HTTP/0.9 Simple-Request receives a full HTTP/1.0 response, which for HTTP/0.9 clients corrupts the body with status line and headers;
- `HEAD` receives an entity body, which section 8.2 prohibits;
- malformed and unsupported requests receive `200 OK` instead of `400` or `501`;
- the response omits the recommended `Date`, `Content-Type`, and `Content-Length` fields and relies solely on connection close to delimit the entity.

RFC 1945 is Informational rather than an Internet Standard, but it is the wire specification this example names and teaches.

**What this PR makes it do.** The handler becomes a small, deliberately conservative RFC 1945 subset (described under Implementation below): it buffers until the request is complete, distinguishes simple from full requests, implements `GET` and `HEAD` correctly, returns meaningful errors, and emits the recommended metadata, while stating plainly in its own docs that it is teaching code, not a production HTTP server.

## Implementation

- Buffer at most 16 KiB while waiting for the request line and headers.
- Recognize a Simple-Request and return only its Simple-Response body.
- Validate HTTP/1.0 request lines, URI octets and escapes, field names, field values, and folded header continuation.
- Implement `GET` and `HEAD`; return `501` for other valid methods and `400` for malformed input.
- Generate RFC-compatible `Date`, `Content-Type`, and `Content-Length` fields.
- Close after one response, matching normal HTTP/1.0 connection behavior.

This remains deliberately compact teaching code. Production applications should use a maintained HTTP server implementation.

## Test plan

Verified on Elixir 1.18.4 / Erlang OTP 27.3.4 against current `main`:

- Split a valid request across TCP sends and verify there is no early response.
- Verify the HTTP/0.9 response contains no HTTP/1.0 status or headers.
- Verify HEAD reports the GET content length but sends no entity body.
- Verify malformed requests return 400.
- Verify invalid request targets, percent escapes, and control bytes in header values return 400, while an absolute URI is accepted.
- Verify an unimplemented valid method returns 501.
- `mix format --check-formatted` and `mix credo --strict` pass.

## Compatibility / risks

This changes only an example that is excluded from the Hex package. The example now uses its default handler state as a request buffer and intentionally handles one request per connection.

### DIFF
--- a/examples/http_hello_world.ex
+++ b/examples/http_hello_world.ex
@@ -1,15 +1,217 @@
 defmodule HTTPHelloWorld do
   @moduledoc """
-  A sample Handler implementation of a simple HTTP Server. Intended to be the
-  simplest thing that can answer a browser request and nothing more. Not even
-  remotely strictly HTTP compliant.
+  A small HTTP/1.0 origin server example implementing a conservative subset of
+  RFC 1945.
+
+  It implements `GET` and `HEAD`, recognizes HTTP/0.9 simple requests, waits for
+  a complete request before responding, and closes each connection after one
+  response. Start it with the default handler options so the initial request
+  buffer is `[]`.
   """
 
   use ThousandIsland.Handler
 
+  @body "Hello, World"
+  @max_request_size 16_384
+  @separators ~c"()<>@,;:\\\"/[]?={} \t"
+  @uri_characters ~c"$-_.+!*'(),;/?:@&="
+
   @impl ThousandIsland.Handler
-  def handle_data(_data, socket, state) do
-    ThousandIsland.Socket.send(socket, "HTTP/1.0 200 OK\r\n\r\nHello, World")
-    {:close, state}
+  def handle_data(data, socket, state) do
+    request = IO.iodata_to_binary([state, data])
+
+    case parse_request(request) do
+      :more ->
+        {:continue, request}
+
+      {:simple, :get} ->
+        ThousandIsland.Socket.send(socket, @body)
+        {:close, request}
+
+      {:full, :get} ->
+        ThousandIsland.Socket.send(socket, response("200 OK", @body, true))
+        {:close, request}
+
+      {:full, :head} ->
+        ThousandIsland.Socket.send(socket, response("200 OK", @body, false))
+        {:close, request}
+
+      {:error, :not_implemented} ->
+        body = "Not Implemented\r\n"
+        ThousandIsland.Socket.send(socket, response("501 Not Implemented", body, true))
+        {:close, request}
+
+      {:error, :bad_request} ->
+        body = "Bad Request\r\n"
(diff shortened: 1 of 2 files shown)


mtrudel:
This completely misses the spirit of the demo, which is to be the simplest possible thing that looks and acts like something someone would recognize as an HTTP request/response pair. The simplicity / naivety of the example is the whole point. 😀

PR #225, opened 2026-08-26
Fix: Omit the TCP_NODELAY default for Unix domain listeners

mpettibone:
Note: I asked Claude to specifically look for spec non-conformance. It found below. Claude helped write the desc, test case and identification of the PR.

## Summary

- Omit Thousand Island's default `nodelay: true` option when the configured listener address is a Unix domain socket, for both the `ip:` and `ifaddr:` forms of `{:local, path}`.
- Apply the change once in the shared `ThousandIsland.Transport.resolve_options/1` helper so TCP and SSL listeners behave identically.
- Document that `nodelay` is conditional and add a regression test that checks the exact options passed to OTP without requiring AF_UNIX support in the test environment.

## Non-conformance

**The rule.** OTP's [`gen_tcp` documentation](https://www.erlang.org/doc/apps/kernel/gen_tcp.html) states it directly: "The option nodelay is a TCP specific option that is not compatible with domain = local." On the backend difference it adds: "This does not actually work for inet_backend = inet either, but in that case the error is simply ignored, which is a bad idea. We have chosen to not ignore this error for inet_backend = socket." The [SSL transport accepts the same underlying listen options](https://www.erlang.org/doc/apps/ssl/ssl.html#type-socket_option). Meanwhile Thousand Island's own documentation supports `ip: {:local, path}` as a listener address, and OTP additionally accepts the `ifaddr: {:local, path}` form.

**What Thousand Island does today.** `ThousandIsland.Transport.resolve_options/1` appends the default `nodelay: true` to every listener's options, Unix domain listeners included, for both the TCP and SSL transports.

**What goes wrong.** Under the legacy inet backend the invalid combination is silently ignored, so it happens to work. Under `inet_backend: :socket`, whether selected per listener or node-wide, the incompatibility is enforced and the listener cannot start at all: on OTP 27.3.4 I observed the listen call fail fatally with a `badarg` exit (the documentation describes the rejection as `{:error, :enotsup}`; the observed shape differs, the substance does not). Thousand Island injects an option OTP declares invalid for an address type Thousand Island declares supported, and the cost is a listener that cannot start on the modern backend.

**What this PR makes it do.** The `nodelay: true` default is omitted when the listener address (via `ip:` or `ifaddr:`) is `{:local, path}`. An explicit `nodelay` value from the caller is passed through unchanged in all cases, and IP listeners keep the existing default.

## Test plan

Verified on Elixir 1.18.4 / Erlang OTP 27.3.4 against current `main`:

- `mix test test/thousand_island/transport_options_test.exs` passes.
- `mix test test/thousand_island/server_test.exs` passes.
- `mix format --check-formatted`, `mix credo --strict`, and `mix dialyzer` pass.
- The new test fails on unpatched `main` and passes with this change.

The new test traces the public transport calls into OTP and asserts that both TCP and SSL omit `nodelay` for the `ip` and `ifaddr` forms of `{:local, path}`, preserve an explicit `nodelay` value, and retain `nodelay: true` for IP listeners. This remains deterministic on hosts where Unix domain sockets cannot be opened. The test uses process-scoped call tracing, runs synchronously, and removes its trace pattern after each transport so other tests are unaffected.

## Compatibility / risks

- IP listeners retain the existing `nodelay: true` default.
- Unix domain listeners no longer receive an inapplicable TCP option, which also unbreaks them under `inet_backend: :socket`.
- An explicitly supplied `nodelay` option is passed through unchanged.

### DIFF
--- a/lib/thousand_island/transport.ex
+++ b/lib/thousand_island/transport.ex
@@ -11,7 +11,6 @@ defmodule ThousandIsland.Transport do
   @required_options [mode: :binary, active: false]
   @default_options [
     backlog: 1024,
-    nodelay: true,
     send_timeout: 30_000,
     send_timeout_close: true,
     reuseaddr: true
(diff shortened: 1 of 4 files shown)


mtrudel:
nice catch!

